PR #14, opened 2017-02-15
Don't assume that id_token response contains the authorization code

ssssam:
I was getting the following error after receiving the ID token from an
instance of the Ipsilon identity server.

    AssertionError: 'undefined' == 'string',
    {"name":"AssertionError","actual":"undefined","expected":"string","operator":"==","message":"'undefined'
    == 'string'","generatedMessage":true}

The issue is in verifying the 'c_hash' field of the response. The code
tried to check it against the 'code' field of the id_token, but this
isn't included in the response from Ipsilon.

I can't find anything in the OpenID Connect Core standard that requires
the authorization code to be included in the ID token response, so it
looks like the client library needs to be fixed to not require that.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -242,7 +242,7 @@ class Client {
     if (params.id_token) {
       promise = Promise.resolve(new TokenSet(params))
         .then(tokenset => this.decryptIdToken(tokenset, 'id_token'))
-        .then(tokenset => this.validateIdToken(tokenset, toCheck.nonce, 'id_token', toCheck.max_age));
+        .then(tokenset => this.validateIdToken(tokenset, toCheck.nonce, 'id_token', toCheck.max_age, params.code));
     }
 
     if (params.code) {
@@ -252,7 +252,7 @@ class Client {
         redirect_uri: redirectUri,
       })
         .then(tokenset => this.decryptIdToken(tokenset, 'id_token'))
-        .then(tokenset => this.validateIdToken(tokenset, toCheck.nonce, 'id_token', toCheck.max_age))
+        .then(tokenset => this.validateIdToken(tokenset, toCheck.nonce, 'id_token', toCheck.max_age, params.code))
         .then((tokenset) => {
           if (params.session_state) tokenset.session_state = params.session_state;
           return tokenset;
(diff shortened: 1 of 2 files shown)


panva:
Hello @ssssam,

i don't see the need to change the way ID Token is validated.

If the received token contains a c_hash or at_hash it means those tokens were issued WITH the ID Token and therefore must be properly validated. 

> I was getting the following error after receiving the ID token from an instance of the Ipsilon identity server.

The ID token should not contain the c_hash claim then. 

If anything I see assertions for c_hash and at_hash presence are missing.

panva:
> The code tried to check it against the 'code' field of the id_token.

This is incorrect, the at_hash get's recalculated with the code property of the whole token set that gets passed in.

ssssam:
Hi, thanks for the feedback. So you think Ipsilon is erroneously setting the c_hash field for the ID token?

I can raise in issue with Ipsilon for sure, but [section 3.3.2.11 of the spec](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitIDTValidation) seems to suggest that it's is up to the implementation whether it returns 'c_hash' even when there's no 'code':

```
If the ID Token is issued from the Authorization Endpoint with a code, which is the case for the response_type values code id_token and code id_token token, this is REQUIRED; otherwise, its inclusion is OPTIONAL. 
```

panva:
No you're right. Few things to fix here, i will do so, together with adding the missing assertions.

puiterwijk:
So, the "code" field in the id_token is *only* required with the Hybrid flow, aka response_mode=id_token or id_token code.
I am seeing response_mode=code, which means authorization code flow.
The fact that I provide the "c_hash" verification value is not required by spec.

panva:
Yup, the issue is that the validation here does not differ between the sources of the id token. It only uses a very simple mechanism (if at_hash, check it) and similar (if c_hash, check it). So the fact that Ipsilon throws in c_hash to the ID Token issued by the Token Endpoint makes the validation fail. I will fix that, and in addition add the missing assertions for presence of these hashes depending on the origin of the Token.

panva:
1.5.3 fixes your immediate issue, 1.6.0 will add assertions for the ..._hash presence where it should be.

ssssam:
Brilliant, 1.5.3 seems to be working for me. Thanks for the quick response! And for writing node-open-client :-)

panva:
@ssssam thank you for bringing the bugs to light!

panva:
@puiterwijk c_hash in a payload of an ID Token from the token_endpoint response is just paperweight :)

PR #21, opened 2017-03-14
prevent revoke unexpected eof error

AVCEngineering:
Typically a revocation call does not have a response body and the current `successHandler` attempts to parse the response body and return it, resulting in `unexpected end of json input` error.

Fix that negates parsing the body if it’s not present.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -587,7 +587,8 @@ class Client {
 
     const body = { token };
     if (hint) body.token_type_hint = hint;
-    return this.authenticatedPost('revocation', { body }, response => JSON.parse(response.body));
+    return this.authenticatedPost('revocation', { body },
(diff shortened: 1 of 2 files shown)


panva:
@AVCEngineering I like this, i guess my servers return 200 and empty json object, but not empty body, for which i would've expected 204 server response code. Regardless, i like the find, but could you add a test for this and possibly return an empty object instead of the response to stay consistent? And i'm not sure the lint will like the ternary in this arrow function.

If not i'll manage myself when i have the time, but won't release a new patch version before then.

AVCEngineering:
@panva thanks for the merge. I was planning to get round to the tests today but you beat me to it 👍

panva:
Happy to!

PR #23, opened 2017-03-20
Change Issuer Assert to OK.

mithun-daa:
When I set the `defaultHttpOptions` to use a port like so:

```
Issuer.defaultHttpOptions = { 
			port: 443,
			agent: tunnel.httpsOverHttp({
				proxy: {
					host: 'myproxy',
					port: 8080
				}
			})
		};
```

The `assert` fails since the `issuer` now has a `:443` tacked on. There might be more elegant ways of doing this but all I could come up with was to use a `startsWith` instead of using the `equal`.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -379,7 +379,7 @@ class Client {
     }
 
     if (payload.iss !== undefined) {
-      assert.equal(this.issuer.issuer, payload.iss, 'unexpected iss value');
+      assert.ok(this.issuer.issuer.startsWith(payload.iss), 'unexpected iss value');
     }
 
     if (payload.iat !== undefined) {


panva:
Thank you for being proactive and trying to help. Unfortunately you're experiencing is not this library's issue - the validation is correct.

1) The port option you are setting is for your http requests, it has nothing to do with the what the your issuer's discovery returns or what the issuer puts in its id_token. > loose the port option, it does nothing imho

2) If the issuer you are discovering identifies (issuer property from well-known/openid-configuration) with the port than the same identifier (including that port!) must be in its tokens.

You can verify the above by taking your discovered `issuer.metadata`, changing the issuer property value to the one with the port you get in the issued tokens and instead of doing a discovery just `new Issuer` with the metadata.

mithun-daa:
The only reason I had to put the port in there is because the discovery was failing if I didn't. I totally understand that this has more to do with the underlying requests. Thanks for the explanation.

panva:
I see, port is then going directly into [http.request options](https://nodejs.org/api/http.html#http_http_request_options_callback). Did you try ditching your agent option and instead running your process like so instead?

```
HTTP_PROXY=http://myproxy:8080 HTTPS_PROXY=http://myproxy:8080 node index.js
```

mithun-daa:
I did, also have these set in my environment variables - doesn't seem to matter.

PR #43, opened 2017-07-17
Allow session key to be specified in passport strategy options

brianGammon:
PR to address the issue I opened earlier. Added 1 new test and all others continue to pass. Thank you and let me know if I missed anything.


panva:
Thank you @brianGammon, well done

PR #55, opened 2017-10-06
send client_id and client secret to google on authorizationCallback

tkabuth:
If the client id and the client secret are not send to google on authorizationCallback, you will get the error: invalid client. So it would be nice if you can add the changes to your repository. Let me know, if I made any mistakes and the changes are not necessary.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -308,6 +308,8 @@ class Client {
         redirect_uri: redirectUri,
         code_verifier: toCheck.code_verifier,
         state: toCheck.state,
+        client_id: this.client_id,
+        client_secret: this.client_secret
       })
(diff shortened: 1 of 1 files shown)


panva:
Hi @tkabuth, thanks for the effort. The different means of authenticating at the OP token_endpoint defined by the specs are already in place. It is just a matter of configuring your client instance right. 

- learn about the various client metadata [here](http://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata)
- discover client auth method [values](http://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication]
- You can see this in the [test suite](https://github.com/panva/node-openid-client/blob/master/test/client/client_instance.test.js#L879-L890)

PR #56, opened 2017-10-10
Add TokenSet support for refresh_expires_in and refresh_expires_at

danmana:
Added support for the non standard token property `refresh_expires_in` (based on similar support for `expires_in`)

Some OAuth2 providers add a `refresh_expires_in` attribute to the token (see [Keycloak](https://issues.jboss.org/browse/KEYCLOAK-760)). 
It behaves the same way `expires_in` does, but it is applied to the `refresh_token`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ package-lock.json
 
 # Optional REPL history
 .node_repl_history
+
+# IDEs and editors
+.idea
--- a/lib/token_set.js
+++ b/lib/token_set.js
@@ -39,6 +39,30 @@ class TokenSet {
   }
 
   /**
+   * @name refresh_expires_in=
+   * @api public
+   */
+  set refresh_expires_in(value) { // eslint-disable-line camelcase
+    this.refresh_expires_at = now() + Number(value);
+  }
+
+  /**
+   * @name refresh_expires_in
+   * @api public
+   */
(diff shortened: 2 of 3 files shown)


panva:
Hello @danmana and thank you for the contribution. 

I'm a little reluctant to add handling for non standardized response parameters but I believe you can extend the TokenSet constructor which is [already exported](https://github.com/panva/node-openid-client/blob/master/lib/index.js#L12) if you want to use this in your code.

Let me know if you need help for that!

danmana:
Thanks :)

I understand if you don't want to add a non standard parameter. This is something I like about this library - how closely you followed the specs.

You can allways merge it later if others show interest in this.

I already covered my use case, thanks.

panva:
Thank you for understanding. For those seeking to extend the TokenSet with custom getters and setters, it's really easy, for instance with the features from this PR.

```js
const { TokenSet } = require('openid-client');

Object.defineProperties(TokenSet.prototype, {
  refresh_expires_in: {
    get() {
      return Math.max.apply(null, [this.refresh_expires_at - now(), 0]);
    },
    set(value) {
      this.refresh_expires_at = now() + Number(value);
    }
  },
  refresh_expired: {
    value() {
      return this.refresh_expires_in === 0;
    }
  }
});
```

Issue #25, opened 2017-03-21
PKCE Support


davidgtonge:
Hi,

Thanks for both this and the node-oidc-provider project!

I'd like to add in support for PKCE.
While this is mainly for public clients, there are use-cases to support it for confidential clients.
I'm thinking of adding a 4th optional argument to this method: 

```javascript
authorizationCallback(redirectUri, parameters, checks) { ... }
// to:
authorizationCallback(redirectUri, parameters, checks, codeVerifier) { ... }
```

No change is needed for the `client.authorizationUrl` method.

What do you think?

panva:
Hi @davidgtonge,

I am much interested in the use-case where this applies to a confidential client that has a kept-secret. I can't think of a situation where PKCE would add any real value to such a client.

What about passing code_verifier as a property of `checks` and if it's present it would be added to the body that goes to the `grant(...)` call.

davidgtonge:
Hi @panva apologies for the delay.
So in an earlier draft of the [FAPI spec](https://bitbucket.org/openid/fapi/commits/8d420fd67b79a1b72c29fac4ea38e6db2b0ff61d?at=master) we had PKCE as required for confidential and public clients. We've now made it only required for public clients, however there is still a "defence in depth" argument for using it for confidential clients. Essentially it provides "Holder of Key" protection for the auth code. So the client is not just proving its identity with its client creds, but also proving that it was the client who started that particular authorization flow.

I also think it's nice to help people understand the specs :-)

I'll do a pull request to add it as an optional property of `checks`.

Thanks

ingoncalves:
The Passport Strategy doesn't take the `code_verifier` param. I'll send a PR with this fix.

PR #59, opened 2017-11-16
Add PKCE support to Passport Strategy.

ingoncalves:
The Passport Strategy didn't take the `code_verifier` param. This PR fixes it.
[#25]

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,6 +399,39 @@ app.get('/auth', passport.authenticate('oidc', [options]));
 app.get('/auth/cb', passport.authenticate('oidc', { successRedirect: '/', failureRedirect: '/login' }));
 ```
 
+### Passport with PKCE
+The Client instance must have the property `token_endpoint_auth_method` setted to `none`.
+See [Proof Key for Code Exchange by OAuth Public Clients][feature-pkce] spec.
+
+
+```js
+const client = new issuer.Client({
+  client_id: 'zELcpfANLqY7Oqas',
+  token_endpoint_auth_method: 'none'
+});
+
+const Strategy = require('openid-client').Strategy;
+
+passport.use('oidc', new Strategy({ client, [params], [passReqToCallback] }, (tokenset, userinfo, done) => {
+    //... same as the previous example
+}));
+
+const verifier = 'foo'; // random PKCE verifier
+
+// start authentication request
+app.get('/auth', passport.authenticate('oidc', {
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -399,6 +399,39 @@ app.get('/auth', passport.authenticate('oidc', [options]));
 app.get('/auth/cb', passport.authenticate('oidc', { successRedirect: '/', failureRedirect: '/login' }));
 ```
 
+### Passport with PKCE
+The Client instance must have the property `token_endpoint_auth_method` setted to `none`.
```
panva: The client instance does not have to have a specific `token_endpoint_auth_method` to use PKCE. As a matter of fact there are OIDC profiles that require the use of PKCE even for confidential authorization_code clients.

lib/passport_strategy.js
```diff
@@ -81,6 +81,7 @@ OpenIDConnectStrategy.prototype.authenticate = function authenticate(req, option
     const state = _.get(session, 'state');
     const maxAge = _.get(session, 'max_age');
     const nonce = _.get(session, 'nonce');
+    const codeVerifier = options.code_verifier;
```
panva: options may be undefined, this is what's causing the test suite to fail.

README.md
```diff
@@ -399,6 +399,39 @@ app.get('/auth', passport.authenticate('oidc', [options]));
...
+}));
+
+// authentication callback
+app.get('/auth/cb', passport.authenticate('oidc', {
+    code_verifier: verifier,
+    successRedirect: '/',
+    failureRedirect: '/login'
+}));
+```
+
```
panva: I feel like the interaction could be made much more simple for the developer. Few options

1) starting the authorization could be made with a verifier. The client can calculate the S256 challenge on its own. There's no need for `yourChallengeFunction` since its standardized, isnt it?
2) PKCE could be automatically used if request is started with code_challenge_method (or simply by setting a strategy instance e.g. `option.usePKCE = true // or "plain", "S256". true would try and look for S256 in issuer configuration, if not found, would use plain, if not exposed, throw?`, verifier would be generated by the strategy and stored in the session just like nonce and states are.

panva:
See #60 for my idea about allowing passport strategy to use PKCE.

ingoncalves:
I appreciate your suggestions and really, all them are valid and necessary. Your implementation in #60 solves this feature much better.

PR #62, opened 2017-11-24
Fix error reporting in issuer field

tayloa45:
We were expecting the wrong value in the issuer field and debugging was a little hard because the error message had the expected and actual values the wrong way around. This should swap them back.


panva:
sweet, thank you for catching this, i went through the rest of the assert.equal calls and found few more.

PR #65, opened 2017-12-11
proxy settings / use request instead of got

panva:
### Proxy settings / Use Request instead of Got
Because of the lightweight nature of got library the client will not use
environment-defined http(s) proxies. In order to have them used you'll need to either provide your own http request
implementation using the provided `httpClient` setter or use the bundled request
one.

Custom implementation:
```js
/*
 * url {String}
 * options {Object}
 * options.headers {Object}
 * options.body {String|Object}
 * options.form {Boolean}
 * options.query {Object}
 * options.timeout {Number}
 * options.retries {Number}
 * options.followRedirect {Boolean}
 */

Issuer.httpClient = {
   get(url, options) {}, // return Promise
   post(url, options) {}, // return Promise
   HTTPError, // used error constructor
};
```

Bundled (and maintained + tested) request implementation after you've added request
to your package.json bundle:

```
npm install request@^2.0.0 --save
```

```js
Issuer.useRequest();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,6 +435,45 @@ Confirm your httpOptions by
 console.log('httpOptions %j', Issuer.defaultHttpOptions);
 ```
 
+### Proxy settings
+Because of the lightweight nature of [got][got-library] library the client will not use
+environment-defined http(s) proxies. In order to have them used you'll need to either provide your own http request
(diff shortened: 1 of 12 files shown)


big8extreme:
thank you very much @panva, you are very reactive 👍

PR #69, opened 2018-01-16
Support for query parameters in defined authorization endpoint

urossmolnik:
Query parameters on `authorization_endpoint` defined on Issuer were ignored when forming `authorizationUrl`.


panva:
Thank you for this!

PR #87, opened 2018-04-07
Set exit code while process exits gracefully

madarche:
cf. https://nodejs.org/api/process.html#process_process_exit_code

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -14,5 +14,5 @@ Issuer.discover(ISSUER).then((issuer) => {
   app.listen(PORT);
 }).catch((err) => {
   console.error(err); // eslint-disable-line no-console
-  process.exit(1);
+  process.exitCode = 1;
 });


panva:
Hi @madarche, what's the point of this change in an example?

madarche:
Hello @panva. At the moment my attention is focused on implementing an OIDC service. And I heavily use and rely on your (high quality) software. That's the "where" reason I'm making PRs on those different projects of yours at the moment.

Now for the "why" reason: I write PRs every time I find something that can be improved in useful or known projects. This has many benefits:

* First, it validates or invalidates my proposed changes. This improves my knowledge.
* Second, it hopefully helps/says thanks to the authors of the projects.
* Third, it hopefully helps others in implementing their own projects. This improves others' knowledge.
* Finally it hopefully helps spreading best practices.

I hope I'm not annoying you too much with those PRs. I intend to propose at least one other PR that will hopefully explain a little more *claims* and *consent* as I've had a hard time with those (and I might also ask some questions in the process if can't make it all alone…). I can stop, just let me know !

With best regards

panva:
Nono, please keep ‘em coming. My only question here is why this specific change, whats the reasoning behind using one and not the other.

panva:
> I intend to propose at least one other PR that will hopefully explain a little more claims and consent as I've had a hard time with those. 

Any PR for docs is most welcome.

madarche:
OK.

As for this specific change: I'm using node-openid-client for testing my OIDC service. And for debugging purposes I added console logs in node-openid-client. But those logs would never got printed with `process.exit(1)`, while with `process.exitCode = 1` those logs had the time to be printed and I had a chance to get the error cause.

panva:
> And for debugging purposes I added console logs in node-openid-client

Where, if there's a throw in the promises this will be logged in line 16

madarche:
> Where, if there's a throw in the promises this will be logged in line 16

On my system, line 16 never has the chance to print anything fast enough, the process is killed too fast with `process.exit(1)`. This is always reproducible on my system with Node.js v8. This behavior is not necessarily reproducible on your system. While with `process.exitCode = 1` the error log is printed to the console and the process die gracefully with the specified exit code when everything is done.

This is what is explained in https://nodejs.org/api/process.html#process_process_exit_code :
> The reason this is problematic is because writes to process.stdout in Node.js are sometimes synchronous and may occur over multiple ticks of the Node.js event loop. Calling process.exit(), however, forces the process to exit before those additional writes to stdout can be performed.

madarche:
I've just updated the comment above. Sorry. Please re-read again.

panva:
fair enough

madarche:
Thanks @panva

PR #91, opened 2018-04-26
fix basic auth to conform with rfc6749 2.3.1

brownbl1:
The purpose of this PR is to fix compliance with [rfc6749 2.3.1](https://tools.ietf.org/html/rfc6749#section-2.3.1), specifically [Appendix B](https://tools.ietf.org/html/rfc6749#appendix-B). 

This library currently implements the Basic authentication scheme, but not the additional requirement of url-encoding both the username and password portion before base64 encoding the whole thing. 

There was a recent [fix](https://github.com/IdentityServer/IdentityServer4/issues/2039) in IdentityServer4 that prompted the discovery of this issue.

A similar implementation can be found [here](https://github.com/lelylan/simple-oauth2/blob/master/lib/encoding.js).

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -797,6 +797,14 @@ class Client {
   }
 
   /**
+   * @name formUrlEncode
+   * @api private
+   */
+  formUrlEncode(value) {
+    return encodeURIComponent(value).replace(/%20/g, '+');
+  }
+
+   /**
    * @name authFor
    * @api private
    */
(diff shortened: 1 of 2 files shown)


panva:
Can you provide a sample (for me) and a test in the test suite for the fixed behavior?

lib/client.js
```diff
@@ -797,6 +797,14 @@ class Client {
   }
 
   /**
+   * @name formUrlEncode
+   * @api private
+   */
+  formUrlEncode(value) {
+    return encodeURIComponent(value).replace(/%20/g, '+');
```
panva: - can `querystring.escape` be used instead of encodeURIComponent?
- what is the point of the replace?

I'd really appreciate tests for this and some references to validate the behavior.

brownbl1:
I was going to mention that this likely affects the provider too. I've seen this all over the place looking at other client libraries.. it's probably an edge case that is not encountered that often.

PR #93, opened 2018-05-16
Add the ability to dynamically copy querystring parameters with passport strategy

major-mann:
Added an additional option to the Passport strategy "copyParams" which will copy all querystring parameters to the authorization URI if it is truthy or, if it is an array, copy any parameters whose names exist in the array.


panva:
Hi @major-mann, I'm going to ahead and close this PR for two reasons

1) it's already something you can achieve before calling `passport.authenticate()`
2) no reference for other passport strategies having it

I would like to thank you for your contribution and welcome any discussion on this or future subjects.

PR #103, opened 2018-07-09
.well-known paths should be treated as suffix

janthoe:
As per https://tools.ietf.org/html/rfc8414#section-3.1 `.well-known` should follow the same pattern as the one stablished used for `.well-known/openid-configuration` in order to keep interoperability.

An example of an existing system using this pattern is Okta, where `.well-known/oauth-authorization-server` is used as a suffix: https://developer.okta.com/docs/api/resources/authorization-servers

ex: https://dev-988291.oktapreview.com/oauth2/ausfmfhs4uinad9qo0h7/.well-known/oauth-authorization-server

### DIFF
--- a/lib/issuer.js
+++ b/lib/issuer.js
@@ -176,7 +176,7 @@ class Issuer {
   static discover(uri) {
     const parsed = url.parse(uri);
 
-    if (parsed.pathname.endsWith(OIDC_DISCOVERY) || parsed.pathname.startsWith(OAUTH2_DISCOVERY)) {
+    if (parsed.pathname.endsWith(OIDC_DISCOVERY) || parsed.pathname.endsWith(OAUTH2_DISCOVERY)) {
       return this.httpClient.get(uri, this.httpOptions())
         .then(expectResponseWithBody(200))
         .then(response => new this(JSON.parse(response.body)))
--- a/test/issuer/discover_issuer.test.js
+++ b/test/issuer/discover_issuer.test.js
@@ -88,7 +88,7 @@ const fail = () => { throw new Error('expected promise to be rejected'); };
     describe('/.well-known/oauth-authorization-server', function () {
       it('accepts and assigns the discovered metadata', function () {
         nock('https://op.example.com', { allowUnmocked: true })
-          .get('/.well-known/oauth-authorization-server')
+          .get('/oauth2/.well-known/oauth-authorization-server')
           .reply(200, {
             authorization_endpoint: 'https://op.example.com/o/oauth2/v2/auth',
             issuer: 'https://op.example.com',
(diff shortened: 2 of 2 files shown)


panva:
Hello,

I’m afraid Okta’s implementation does not follow the final rfc but one of the drafts where the document location was absolutely equal. 

If you read carefully the section linked and then section 5 you’ll see the differences in resolving the document location in cases where the issuer contains a path component.

panva:
From section 3.1

> If the issuer identifier value contains a path component, any terminating `/` MUST be removed before inserting `/.well-known/` and the well-known URI suffix between the host component and the path component.  The client would make the following request when the issuer identifier is `https://example.com/issuer1` and the well-known URI suffix is `oauth-authorization-server` to obtain the metadata, since the issuer identifier contains a path component:

     GET /.well-known/oauth-authorization-server/issuer1 HTTP/1.1
     Host: example.com

From section 5

> The algorithm for transforming the issuer identifier to an authorization server metadata location defined in Section 3 is equivalent to the corresponding transformation defined in Section 4 of "OpenID Connect Discovery 1.0", **provided that the issuer identifier contains no path component.  However, they are different when there is a path component**, because OpenID Connect Discovery 1.0 specifies that the well-known URI string is appended to the issuer identifier (e.g., `https://example.com/issuer1/.well-known/openid-configuration`), whereas this specification specifies that the well-known URI string is inserted before the path component of the issuer identifier (e.g., `https://example.com/.well-known/openid-configuration/issuer1`).

janthoe:
I see, the changes `draft-09` make no sense at all but your library is doing the right thing. Thanks for your time.

panva:
FYI, 

it seems okta returns okay results to both

- https://dev-988291.oktapreview.com/oauth2/ausfmfhs4uinad9qo0h7/.well-known/oauth-authorization-server
- https://dev-988291.oktapreview.com/oauth2/ausfmfhs4uinad9qo0h7/.well-known/openid-configuration

And for some unknown reason to me they differ. (e.g. oauth-authorization-server is missing userinfo_endpoint).

You should just pass `https://dev-988291.oktapreview.com/oauth2/ausfmfhs4uinad9qo0h7/.well-known/openid-configuration` to Issuer.discover and you're good to go - in that case `oauth-authorization-server` won't be looked up.

panva:
And 0c8a8f242effe2e444db36271de68b4b71ea8489 covers custom well-known suffix as well as your use-case (to be released, end of today).

janthoe:
@panva Thanks for the info.
Okta has a setup where they only have one authorization server per tenant that allows user-info requests. All other authorization servers in the tenant will reject reject requests with `openid` scope. According to Okta support that is the intended design 🤷‍♂️

janthoe:
@panva: Thanks for the changes in 0c8a8f2 🙇

panva:
> @panva Thanks for the info.
Okta has a setup where they only have one authorization server per tenant that allows user-info requests. All other authorization servers in the tenant will reject reject requests with `openid` scope. According to Okta support that is the intended design 🤷‍♂️ 

That's fine, but their location should still be fixed :)

I assume there will be more IdPs that get it wrong, hence 0c8a8f2

Issue #120, opened 2018-09-15
Sending state on auth code token grant no longer recommended


davidgtonge:
This library currently sends the state param to the token endpoint when doing an auth code token grant. This was one of the mitigations to "mix up" attacks. However it is no longer recommended as per: https://datatracker.ietf.org/doc/draft-ietf-oauth-security-topics/

We also came across behaviour where some OPs rejected calls to the token endpoint when `state` was included. While those OPs are probably in the wrong, I don't see a need for this param to be sent any more (the library already supports PKCE, nonce and s_hash verification. 

Further discussion about this issue is also here:
https://bitbucket.org/openid/fapi/issues/173/mix-up-mitigation-defence-in-depth

davidgtonge:
Thanks @panva - I'm continually amazed at how responsive you are.
I'll have a go at opening a PR with the `s_hash` check from the token endpoint

panva:
If you don’t mind, please hit me up via email, i’d like to extend the auth method list with mtls and would like to do so without disrupting your existing setup, altho i have a pretty good idea of what it is (none + http options with client cert) i’d like to know for sure the specific nuances that make it usable in fapi + ob right now already.

PR #122, opened 2018-09-15
feat: add support for s_hash verification at token endpoint

davidgtonge:
This is required in the OpenID Foundation's FAPI profile if s_hash hasn’t
been returned from the authorization endpoint.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -378,7 +378,7 @@ class Client {
         code_verifier: checks.code_verifier,
       })
         .then(tokenset => this.decryptIdToken(tokenset))
-        .then(tokenset => this.validateIdToken(tokenset, checks.nonce, 'token', checks.max_age))
+        .then(tokenset => this.validateIdToken(tokenset, checks.nonce, 'token', checks.max_age, checks.state))
         .then((tokenset) => {
           if (params.session_state) tokenset.session_state = params.session_state;
           return tokenset;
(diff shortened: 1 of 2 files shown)


davidgtonge:
As discussed in #120

davidgtonge:
Oops - just pushed the tls_client_auth to the wrong branch - will fix that and add the assertion.

Flag for the profile sounds good.

panva:
I’m travelling atm. Will take a look at all of this tomorrow evening.

davidgtonge:
sounds good - I also probably wont get a chance to finish these off until then.

davidgtonge:
@panva so I've pushed a commit that checks the presence of s_hash, however how do you want to define the different profiles. 

Should it be something like `clientInstance.PROFILE = Issuer.PROFILES.FAPI2`
Or just clientInstance.PROFILE = "FAPI2"

Also this is the relevant clause:
> shall include state hash, s_hash, in the ID Token to protect the state value if the client supplied a value for state. s_hash may be omitted from the ID Token returned from the Token Endpoint when s_hash is present in the ID Token returned from the Authorization Endpoint;

I think it can be simplified (and made clearer)
FAPI Part 2 requires hybrid mode, therefore there is always an id token returned in the auth response, therefore if state is provided, the s_hash must be present in that id_token. 

I'm not sure if there is any benefit from it being included in the id token from the token endpoint.

panva:
@davidgtonge this is already looking good, but let's see what differences does FAPI/OB/pure OIDC make in the mtls case to nail down the profile option.

The options I see

1) define the profile constants and export them similar to your suggestion above
2) define a new set of constructors FAPIIssuer, FAPIClient (or similar) which have these extra checks enabled by default, for instance - they require s_hash from the authorization endpoint, they only allow hybrid response_type to be used and they require that checks.response_type be provided so that the response parameter presence gets verified.

Furthermore support for JARM signed/encrypted authorization response parsing is needed.

panva:
@davidgtonge does FAPI make the use of state parameter required?

davidgtonge:
> does FAPI make the use of state parameter required?

Not at the moment. The tricky thing is that FAPI is trying to walk a line between OAuth 2.0 and OIDC and its bringing some confusion. For example in OIDC with a hybrid flow, the `nonce` should be enough. But  we want FAPI to be able to support https://tools.ietf.org/html/draft-ietf-oauth-jwsreq-16 and JARM. In that case, state should be required. 

I'll have a think about the differences required for MTLS between the different profiles (hopefully there aren't any).

panva:
@davidgtonge seeing that FAPI requires hybrid to be used, is adding the validation to token endpoint necessary for your case? The profile is something i'll get to once i get per-endpoint http options and mtls going.

Issue #127, opened 2018-10-05
Missing check if Error.captureStackTrace is undefined


Eblax:
I'm using this module in an Ionic mobile project. This works fine in Chrome and on Android, but it would not run correctly on iPhone and Firefox. I tracked the issue down to a missing check if `Error.captureStackTrace` is defined in `lib\open_id_connect_error.js`, which would cause `Issuer.discover` to stop running.

For a quick workaround I added `if(Error && !Error['captureStackTrace']) { Error['captureStackTrace'] = function() {} }` before the first call to `Issuer.discover`.

I see most other node modules in my project do an `if (Error.captureStackTrace)` before running it so maybe that could be an idea?

panva:
Let me know how it goes, if it comes down to just this and you don't hit any other issues, I'll merge the PR.

Eblax:
Seems like other modules required by this one (e.g. _got_) have the same issue, so feel free to disregard the pull request. The error-polyfill had some issues on its own ('Unexpected frame by generating stack.') so I'll stick with my workaround for the time being. Thanks for the quick reply!

panva:
👌

panva:
FWIW this seems to be the polyfill you should use to actually get stack traces.

```js
var captureStackTrace = Error.captureStackTrace
if (captureStackTrace === undefined) {
  captureStackTrace = function captureStackTrace (error) {
    var container = new Error()

    defineProperty(error, 'stack', {
      configurable: true,
      get: function getStack () {
        var stack = container.stack

        // Replace property with value for faster future accesses.
        defineProperty(this, 'stack', {
          configurable: true,
          value: stack,
          writable: true
        })

        return stack
      },
      set: function setStack (stack) {
        defineProperty(error, 'stack', {
          configurable: true,
          value: stack,
          writable: true
        })
      }
    })
  }
}
```

Credit: https://github.com/JsCommunity/make-error/blob/f3086a7f33ad1451e474fbcace5698275fa27eda/index.js#L12-L40

Eblax:
Thanks, that seems to work like a charm (after replacing defineProperty with Object.defineProperty)
I'll use that to make my workaround a bit less workaroundy :)

PR #128, opened 2018-10-05
Added check for Error.captureStackTrace

Eblax:
As mentioned in #127


panva:
closing based on #127 feedback

Issue #31, opened 2017-04-27
more logging / debug options


JoshBarr:
Hey @panva, 

Thanks for making a great library. This makes doing OIDC in Node really easy. 

I was wondering, would you consider accepting a PR to add some logging? The API I was considering was this:

```js
/*
interface ILogger {
  log(message: string)
  warn(message: string)
  info(message: string)
  error(message: string)
}
*/

const issuer = new Issuer({
   logger: console,  /* implements ILogger */
});

// Would inherit logger from issuer, but could also pass logger into the client
const client = new issuer.Client({ 
   logger:  console,  /* implements ILogger */
});
```

My use case is a container cluster which I monitor via CloudWatch. It's not easy to tell what errors we're getting from `got`, particularly when proxies or redirects are involved. If you're happy with the API, I can write the code/tests.

panva:
Hello @JoshBarr, i think this would be a great addition, even better if you could contribute it.

my $.02 towards the API solution direction,
- do we need issuer / client instance specific loggers?
- do we need Issuer / Client specific loggers? > can't we do with one logger setting (similar to defaultHttpOptions set on Issuer) This would greatly reduce the efforts and keep the API simple and understandable.
- can the default logger state be disabled / not provided?
- is your problem error specific? Would you consider Issuer/Client being an EventEmitter?

JoshBarr:
You're right, we probably don't need client-specific loggers. Something like this?

```js
const Issuer = require('openid-client').Issuer;
Issuer.logger = console;
```

Making the Iss/Client into EventEmitters would actually be really neat – certainly would make logging easier, as we can just attach the logger to the events.

panva:
This already looks the right amount of "simple" I shoot for ;) Let me get back to you tomorrow, I will give it some thought whether to do emits or a logger.

panva:
@JoshBarr what do you think of a simple inclusion of [debug](https://github.com/visionmedia/debug) instead?

JoshBarr:
Debug would be fine for our use case - we only need to log some internals for troubleshooting purposes.

panva:
Sounds good!

panva:
Feel free to open PR, 'till then i will be closing non-Issues.

vancouverwill:
hi @panva I needed to add in some logs into a fork of this repo so we could see why it was failing for us, i.e. which stage and for what reason and thought it could be useful for others.

It looks like this issue is closed for a while but I still think it useful. 

[...]

PR #130, opened 2018-10-14
feat: add optional debug logs

vancouverwill:
#31 

WIP

todo:
* log other endpoints
* update readme to show how to use debugger

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -2,9 +2,10 @@ const util = require('util');
 const assert = require('assert');
 const stdhttp = require('http');
 const crypto = require('crypto');
-const querystring = require('querystring');
 const url = require('url');
+const querystring = require('querystring');
 
+const debug = require('debug')('openidclient');
 const jose = require('node-jose');
 const uuid = require('uuid/v4');
 const base64url = require('base64url');
@@ -288,7 +289,7 @@ class Client {
     let hint = params.id_token_hint;
 
     if (hint instanceof TokenSet) {
-      assert(hint.id_token, 'id_token not present in TokenSet');
+      assert(hint.id_token, 'endSessionUrl() id_token not present in TokenSet');
       hint = hint.id_token;
     }
 
@@ -359,6 +360,7 @@ class Client {
    * @api public
    */
   authorizationCallback(redirectUri, parameters, checks = {}) {
+    this.issuer.logger({ message: 'SSO begin authorizationCallback' });
     const params = _.pick(parameters, CALLBACK_PROPERTIES);
 
     if (this.default_max_age && !checks.max_age) checks.max_age = this.default_max_age;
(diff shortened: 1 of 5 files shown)


lib/client.js
```diff
@@ -288,7 +289,7 @@ class Client {
     let hint = params.id_token_hint;
 
     if (hint instanceof TokenSet) {
-      assert(hint.id_token, 'id_token not present in TokenSet');
+      assert(hint.id_token, 'endSessionUrl() id_token not present in TokenSet');
```
vancouverwill: updated the error message here so it's source was transparent. I have done the same for other places where id_token is not set

panva:
I’m not fond of the logger option, sorry. “debug” module at most

vancouverwill:
> I’m not fond of the logger option, sorry. “debug” module at most

Thanks for looking and fair enough. Would you mind explaining why you don't like the logger option? **debug** library only supports one style of log format and we need json formatted logs in production for log search.
I looked and thought about using debug as the only logger for a while this weekend as we do need some transparency around what is going on within **node-openid-client** but using debug is an rigid approach to logs and will only suit certain use cases. This seemed a good compromise, to default to **debug** but to allow clients of **node-openid-client** to use their own logging system.

vancouverwill:
> First, debug capabilities must not be confused with logging, i believe none of the request/response objects should ever reach a log entry unmodified, there's PII, bearer tokens and client secret credentials all over the place.

Yeah good point, the logs I put in would certainly not be safe in a production environment. Creating events would be a lotpre flexible solution for most people.
> id_token not present in TokenSet if the IdP you're using isn't consistent with returning ID Tokens i.e. after refreshes, they ought to let you know and you shouldn't blindly call tokenset claims. It is, afterall, a really stupid helper that you can do yourself IF you have an id token.
The reason I updated these error messages is each one had sent content so made it hard to see error came from. Unique errors do help, and in authorize callback the grant and token validation errors are all caught with same catch so still bit ambiguous as to what happened. What do you think about an error getting thrown in the grant request  so you at least know that is what went wrong ?
Thanks for your time 👌

panva:
> The reason I updated these error messages is each one had sent content so made it hard to see error came from. Unique errors do help, and in authorize callback the grant and token validation errors are all caught with same catch so still bit ambiguous as to what happened. What do you think about an error getting thrown in the grant request so you at least know that is what went wrong ?

I'm sorry I don't follow, the problem is?

vancouverwill:
> > The reason I updated these error messages is each one had sent content so made it hard to see error came from. Unique errors do help, and in authorize callback the grant and token validation errors are all caught with same catch so still bit ambiguous as to what happened. What do you think about an error getting thrown in the grant request so you at least know that is what went wrong ?
> 
> I'm sorry I don't follow, the problem is?

ahh sorry, I guess that is two different ideas
1 - the same error message `id_token not present in TokenSet` is used as an error message in several places so it is hard to know the source of the error
2 - inside the `authorizeCallback()` there is a lot of stuff going on with grant(), decryptIdToken() and validateIdToken() which is why I needed to update the grant function so I could see where the failure has happened. Would you be open to consider to having a different error getting thrown if the grant fails?

PR #135, opened 2018-11-09
Passport querystring passthrough

major-mann:
Added a constructor parameter to the passport strategy which allows querystring values to be passed along with the client authorization redirect.

### DIFF
--- a/lib/passport_strategy.js
+++ b/lib/passport_strategy.js
@@ -32,6 +32,7 @@ function OpenIDConnectStrategy({
   passReqToCallback = false,
   sessionKey,
   usePKCE = false,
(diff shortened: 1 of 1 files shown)


major-mann:
Need to revisit the tests

panva:
didn't you attempt #93 already? You can achieve this without any extra code.

major-mann:
The purpose here is to be able to pass a token as part of a link (Not a static value at configuration time). This allows the login form to be skipped when, for example, sending out a mail campaign.

For example: http://www.example.com/login?auto-login-token=abc123

panva:
See #136.

PR #136, opened 2018-11-09
Allow pass through of defined querystring parameters in the passport strategy

major-mann:
---
name: Feature addition
type: Feature addition
about: Allow querystring parameters to be passed through when using passport strategy
---

This change allows the definition of querystring parameters that should be passed along to the server during authorization. These parameters will be attached to the redirection URI.

### DIFF
--- a/lib/passport_strategy.js
+++ b/lib/passport_strategy.js
@@ -22,6 +22,22 @@ function verified(err, user, info = {}) {
   }
 }
 
+/** Reads through any supplied pass through keys and copies their values from the querystring */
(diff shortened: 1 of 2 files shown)


panva:
ad https://github.com/panva/node-openid-client/pull/135#issuecomment-437361103

As pointed in #93 you can already achieve that like so

```js
app.get('/auth', function (req, res) {
  const additionalParams = {}; // assign your passthroughs to additionalParams here.
  passport.authenticate('oidc', params)(req, res);
});
```

PR #149, opened 2019-02-13
feat: support extra claims in JWT client_assertion

apparentvisuals:
OpenID allows extra claims to be added to client_assertion as part of private_key_jwt and client_secret_jwt. Added an optional and backward compatible way to insert extra claims into the client_assertion during a grant request with unit tests.

### DIFF
--- a/lib/helpers/consts.js
+++ b/lib/helpers/consts.js
@@ -37,6 +37,7 @@ const CALLBACK_PROPERTIES = [
   'state',
   'token_type',
   'session_state',
+  'resource',
 ];
 
 const DEFAULT_HTTP_OPTIONS = {
--- a/lib/client.js
+++ b/lib/client.js
@@ -948,22 +948,26 @@ class Client {
       case 'private_key_jwt':
       case 'client_secret_jwt': {
         const timestamp = now();
-        extraClaims = extraClaims || {};
-        return this.createSign(endpoint).then(sign => sign.update(JSON.stringify(Object.assign({}, {
+        const allClaims = JSON.stringify(Object.assign({}, extraClaims || {}, {
           iat: timestamp,
           exp: timestamp + 60,
           jti: random(),
           iss: this.client_id,
           sub: this.client_id,
           aud: this.issuer[`${endpoint}_endpoint`],
-        }, extraClaims))).final().then((client_assertion) => { // eslint-disable-line camelcase, arrow-body-style
-          return {
-            body: {
-              client_assertion,
-              client_assertion_type: 'urn:ietf:params:oauth:client-assertion-type:jwt-bearer',
-            },
-          };
         }));
+
+        return this.createSign(endpoint).then((sign) => { // eslint-disable-line arrow-body-style
+          // eslint-disable-next-line camelcase, arrow-body-style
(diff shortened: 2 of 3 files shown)


panva:
Please swap the merge order of extra vs well defined. Whatever is provided extra should not overwrite the conform implementation.

panva:
@apparentvisuals Just checking in, i haven't forgotten about this, just now planning to revive the work on 3.x (due in April with node 12 release) and i'll include this there.

lib/helpers/consts.js
```diff
@@ -37,6 +37,7 @@ const CALLBACK_PROPERTIES = [
   'state',
   'token_type',
   'session_state',
+  'resource',
```
panva: `resource` is not a IANA registered authorization response property, please remove this.

apparentvisuals: Sorry, this wasn't suppose to be part of this pull request. Would you accept a separate pull request for this to support resource indicators https://tools.ietf.org/html/draft-ietf-oauth-resource-indicators-01 ?

panva: just callback wouldn't be full support. Let me take a note and include this in the 3.x due in a month, i'll do a full support there.

apparentvisuals: Sounds good. Do you still need the pull request for extra claims in JWT client assertion? If so I'll branch the change and re-upload a pull request.

panva: no, you can close this one. It's on my list to do something about it.

panva:
@apparentvisuals can you explain what you need to add to the JWT client assertions?

apparentvisuals:
We add some extra service specific metadata as part of the client assertion claims for generating slightly different token claims. Since in OpenID it is allowed to add extra claims to the JWT there wasn't any need to implement every single one as a custom grant type.

panva:
I can understand extra request body properties but please clarify what’s the point of extending the client assertion

apparentvisuals:
When used during client_credentials grant or custom grant types I see it similar to how Request Object works for authorization_code, the values could be in the request body, but inside the assertion the entire value set becomes signed by the private_ket_jwt.

panva:
I still fail to see the point given that the assertion itself is a one-time use.

apparentvisuals:
I agree, there may not be much difference between request body and inside the assertion JWT, I needed it because I used some grant types against existing idp that used it that way, and openid connect core specifically allows custom claims to be part of the assertion I didn't see any reason to not support this in a generic openid client for flexibility of the API.

panva:
The point behind the statements about other claims is mostly extensibility by future profiles and specifications. Just sayin’

It will be part of 3.x nontheless but i would stay away from an IdP that’s not interoperable and seems to make up its own “profiles”.

pr4xx:
Is this already implemented? We really need a way to add extra body params to the code exchange.

panva:
[yes](https://github.com/panva/node-openid-client/tree/master/docs#clientcallbackredirecturi-parameters-checks-extras), see the extras param.

pr4xx:
Thanks! Is there any way to add these extra params to the passport strategy?

PR #164, opened 2019-04-30
fix: increase http timeout to 10s

sami-sweng:
Hi @panva,

Thank you for your quick reactions and fix yesterday!

So as I told you, I was making the PR because of a server that randomly timedout when calling it's key route. So here is the PR that we will use in prod to make 100% sure that this scenario is not happening. Since we have a working cache now it should almost never be used, but better safe than sorry.


panva:
You can [customize the timeout](https://github.com/panva/node-openid-client#changing-http-request-defaults) and other request object details in your own code base.

PR #184, opened 2019-09-05
feat: Add Typescript definitions

theogravity:
I've handcrafted TS defs that is based on the API documentation and analysis of the project files.

We have had no problems using it in a project we're working on so far.

I'm unsure if you've worked with Typescript or not, but I've done my best to make sure that for return types where a class is involved, that an interface is returned instead.

This allows for the creation of a mocked class when writing unit tests, vs doing class definition overrides. 

An exception case would be `TokenSet` as it doesn't do anything complicated and is pretty much a plain object under the hood.


panva:
Whoa! @theogravity this is great, I will go through the PR to make sure nothing was missed soon.

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    interval: number
+    user_code: string
+    device_code: string
+    verification_uri: string
+    verification_uri_complete: string
+    expires_in: number
+  }
+
+  export class DeviceFlowHandle implements IDeviceFlowHandle {
+    constructor (params : IDeviceFlowHandleParams)
```
panva: remove the constructor from types please, the class is not intended to be constructed manually

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    response: string
+    maxAge: number
+  }
+
+  export interface IDeviceFlowHandle {
+    poll() : Promise<TokenSet>
+    expired() : boolean
+    expires_at: number
+    client: IClient
+    maxAge: number
```
panva: remove `maxAge`

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    exchangeBody: object
```
panva: remove `exchangeBody`

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    clientAssertionPayload: object
```
panva: remove `clientAssertionPayload`

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    response: string
```
panva: remove `response`

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    interval: number
```
panva: remove `interval`

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    user_code: string
+    // tslint:disable-next-line:variable-name
+    device_code: string
+    // tslint:disable-next-line:variable-name
+    verification_uri: string
+    // tslint:disable-next-line:variable-name
+    verification_uri_complete: string
+    // tslint:disable-next-line:variable-name
+    expires_in: number
+  }
```
panva: same removals as above, in the `DeviceFlowHandle` i don't see the point duplicating everything from the Interface.

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+
+    [key: string]: any
+  }
+
+  export interface IMetadataMtlsEndpointAliases {
+    token_endpoint: string
+    userinfo_endpoint: string
+    revocation_endpoint: string
+    introspection_endpoint: string
+  }
```
panva: add `device_authorization_endpoint`

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    token_endpoint_auth_signing_alg_values_supported: string
+    introspection_endpoint_auth_methods_supported: string
+    introspection_endpoint_auth_signing_alg_values_supported: string
+    revocation_endpoint_auth_methods_supported: string
+    revocation_endpoint_auth_signing_alg_values_supported: string
+    request_object_signing_alg_values_supported: string
+    mtls_endpoint_aliases: IMetadataMtlsEndpointAliases
+
+    [key: string]: any
+  }
```
panva: add `device_authorization_endpoint`

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    /**
+     * space-separated scope(s) used for the authentication request
+     */
+    scope?: string
+    expires_in?: number
+    expires_at?: number
+    /**
+     * State value passed in the authentication request
+     */
+    session_state?: string | object
```
panva: never an object.

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    // tslint:disable-next-line:variable-name
+    id_token?: string
+    // tslint:disable-next-line:variable-name
+    refresh_token?: string
+    // tslint:disable-next-line:variable-name
+    expires_in?: number
+    // tslint:disable-next-line:variable-name
+    expires_at?: number
+    // tslint:disable-next-line:variable-name
+    session_state?: string | object
```
panva: never an object

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    revocation_endpoint: string
+    end_session_endpoint: string
+    registration_endpoint: string
+    token_endpoint_auth_methods_supported: string
+    token_endpoint_auth_signing_alg_values_supported: string
+    introspection_endpoint_auth_methods_supported: string
+    introspection_endpoint_auth_signing_alg_values_supported: string
+    revocation_endpoint_auth_methods_supported: string
+    revocation_endpoint_auth_signing_alg_values_supported: string
+    request_object_signing_alg_values_supported: string
```
panva: any property ending with `_supported` is a `string[]`, not a string.

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+
+  /**
+   * Encapsulates a discovered or instantiated OpenID Connect Issuer (Issuer), Identity Provider (IdP),
+   * Authorization Server (AS) and its metadata.
+   */
+  export interface IIssuer {
+    /**
+     * Returns the <Client> class tied to this issuer.
+     */
+    Client: { new (metadata: IClientMetadata, jwks?: object) }
```
panva: dtto as `jwks` above.

panva:
@ulrichb @YangusKhan @RMHonor as the three contributors to the DefinitelyTyped types, i would appreciate your review here as well.

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+ */
+declare module 'openid-client' {
+  import { JWKS } from '@panva/jose'
+  import { IncomingMessage } from 'http'
+  import { Http2ServerRequest } from 'http2'
+
+  /**
+   * @see https://medium.com/@darutk/diagrams-of-all-the-openid-connect-flows-6968e3990660
+   */
+  enum AuthResponseType {
```
RMHonor: If we're defining ambient enums in the declaration file, they'll also need to exist in the javascript to be useable

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    CODE = 'code',
+    TOKEN = 'token',
+    ID_TOKEN = 'id_token',
+  }
+
+  /**
+   * @see https://github.com/panva/node-openid-client/blob/master/docs/README.md#client-authentication-methods
+   */
+  enum TokenAuthMethod {
```
RMHonor: If we're defining ambient enums in the declaration file, they'll also need to exist in the javascript to be useable

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    request_object_encryption_enc?: string
+    token_endpoint_auth_method?: TokenAuthMethod
+    introspection_endpoint_auth_method?: TokenAuthMethod
+    revocation_endpoint_auth_method?: TokenAuthMethod
+    token_endpoint_auth_signing_alg?: string
+    introspection_endpoint_auth_signing_alg?: string
+    revocation_endpoint_auth_signing_alg?: string
+    tls_client_certificate_bound_access_tokens?: boolean
+
+    [key: string]: any
```
RMHonor: `unknown` is a safer type to use here, so the developer has to explicitly cast it to what they believe the type should be

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+  }
+
+  export interface IAuthorizationUrlParams {
+    redirect_uri?: string
+    response_type?: string
+    scope?: string,
+    [key: string]: any
```
RMHonor: `unknown` is a safer type to use here, so the developer has to explicitly cast it to what they believe the type should be

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    post_logout_redirect_uri?: string
+    state?: string
+  }
+
+  export type CallbackParamsType = {
+    state?: string
+    response_type?: string
+    nonce?: string
+    code_verifier?: string
+    [key: string]: any
```
RMHonor: `unknown` is a safer type to use here, so the developer has to explicitly cast it to what they believe the type should be

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    /**
+     * extra client assertion payload parameters to be sent as part of a client JWT assertion.
+     * This is only used when the client's token_endpoint_auth_method is either client_secret_jwt or private_key_jwt.
+     */
+    clientAssertionPayload?: object
+  }
+
+  export interface IGrantBody {
+    grant_type: string
+    [key: string]: any
```
RMHonor: `unknown` is a safer type to use here, so the developer has to explicitly cast it to what they believe the type should be

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+  }
+
+  export interface IRequestObjectPayload {
+    client_id?: string
+    iss?: string
+    aud?: string
+    iat?: number
+    exp?: number
+    jti?: string
+    [key: string]: any
```
RMHonor: `unknown` is a safer type to use here, so the developer has to explicitly cast it to what they believe the type should be

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    /**
+     * Initial Access Token to use as a Bearer token during the registration call.
+     */
+    initialAccessToken?: string
+  }
+
+  export interface IDeviceAuthParameters {
+    client_id?: string
+    scope?: string
+    [key: string]: any
```
RMHonor: `unknown` is a safer type to use here, so the developer has to explicitly cast it to what they believe the type should be

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+     */
+    requestObject(payload: IRequestObjectPayload): Promise<string>
+
+    /**
+     * Starts a Device Authorization Request at the issuer's device_authorization_endpoint and returns a handle
+     * for subsequent Device Access Token Request polling.
+     */
+    deviceAuthorization(parameters?: IDeviceAuthParameters, extras?: IDeviceAuthExtras): Promise<IDeviceFlowHandle>
+
+    [key: string]: any
```
RMHonor: Are you sure we want this extension here? We shouldn't be assuming there are additional properties on a client

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+    token_endpoint_auth_methods_supported: string
+    token_endpoint_auth_signing_alg_values_supported: string
+    introspection_endpoint_auth_methods_supported: string
+    introspection_endpoint_auth_signing_alg_values_supported: string
+    revocation_endpoint_auth_methods_supported: string
+    revocation_endpoint_auth_signing_alg_values_supported: string
+    request_object_signing_alg_values_supported: string
+    mtls_endpoint_aliases: IMetadataMtlsEndpointAliases
+
+    [key: string]: any
```
RMHonor: Ditto with `unknown`

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+     */
+    metadata?: IIssuerMetadata
+
+    /**
+     * Returns the issuer's jwks_uri keys as a @panva/jose parsed JWKS.Keystore.
+     * @param forceReload
+     */
+    keystore (forceReload?: boolean): Promise<JWKS.KeyStore>
+
+    [key: string]: any
```
RMHonor: As above, should an `Issuer` allow additional types like this?

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+     */
+    scope?: string
+    expires_in?: number
+    expires_at?: number
+    /**
+     * State value passed in the authentication request
+     */
+    session_state?: string | object
+
+    [key: string]: any
```
RMHonor: `unknown` again

RMHonor: on second thought, should we allow this to be extensible at all? I can't think of an instance where a `TokenSet` will have additional properties, @panva may be better placed to comment

big-kahuna-burger:
Can we also have `Strategy` exported for the sake of completeness?

lib/index.d.ts
RMHonor: May be more suitable to use a string literal if we want to restrict the set of auth types without needing to access an existing enum in JS

lib/index.d.ts
panva: Anything returned by the token response is available actually. So yeah. But as you suggest above, unknown as a type.

lib/index.d.ts
panva: Agreed, ergo please change all enums to `type Foo = 'a' | 'b' | 'c'` instead

panva:
I'd like someone to write .ts test file using the types so that we may have a CI job verifying them. Any takers?

big-kahuna-burger:
> I'd like someone to write .ts test file using the types so that we may have a CI job verifying them. Any takers?

@panva I will take it (via `tsd`). Expect PR later in the evening.

Bessonov:
Great! While it's better to have definitions instead of not to have definitions at all... wouldn't it better to convert the project to typescript? What do you think @panva ? Maybe step by step.

> I'd like someone to write .ts test file using the types so that we may have a CI job verifying them.

I'm not sure that this would help to check definitions. AFAIK you will check fictional*1 definitions with fictional*2 tests.

*1 fictional, because they can be not related to concrete implementation.
*2 fictional, because it doesn't make much sense to create high coverage tests just for checking typings.

Maybe it's better to convert tests to ts as a first step?

panva:
Let's park the tests for now until we're satisfied with the state of the definition.

@Bessonov exposing types is as far as I'm willing to go at this point in time.

lib/index.d.ts
```diff
@@ -0,0 +1,528 @@
...
+     * Starts a Device Authorization Request at the issuer's device_authorization_endpoint and returns a handle
+     * for subsequent Device Access Token Request polling.
+     */
+    deviceAuthorization(parameters?: IDeviceAuthParameters, extras?: IDeviceAuthExtras): Promise<IDeviceFlowHandle>
+
+    [key: string]: any
+  }
+
+  export class Client implements IClient {
+    constructor (metadata: IClientMetadata, jwks?: object)
```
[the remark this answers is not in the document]
theogravity: sure

theogravity:
Thank you everyone for the quick turnaround!

I've addressed the most recent comments - apply `JSONWebKeySet`, and remove the `DeviceFlowHandle` def.

lib/index.d.ts
```diff
@@ -186,11 +186,11 @@ declare module 'openid-client' {
      * JWK Set formatted object with private keys used for signing client assertions or decrypting responses.
      * When neither jwks_uri or jwks is present in metadata the key's public parts will be registered as jwks.
      */
-    jwks?: object
+    jwks: JSONWebKeySet
     /**
      * Initial Access Token to use as a Bearer token during the registration call.
      */
-    initialAccessToken?: string
+    initialAccessToken: string
```
panva: this is still optional

lib/index.d.ts
```diff
@@ -186,11 +186,11 @@ declare module 'openid-client' {
      * JWK Set formatted object with private keys used for signing client assertions or decrypting responses.
      * When neither jwks_uri or jwks is present in metadata the key's public parts will be registered as jwks.
      */
-    jwks?: object
+    jwks: JSONWebKeySet
```
panva: this is still optional

panva:
@theogravity thank!

Re-iterating on the rest of missing pieces

> In addition to those we neeed to add [generators](https://github.com/panva/node-openid-client/tree/master/docs#generators) and the [customization](https://github.com/panva/node-openid-client/tree/master/docs#customizing) functionality, some of this is already part of the community maintained [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/openid-client/index.d.ts).

I appreciate the effort you've put into this.

theogravity:
Todo:
-`Strategy`
-`Errors`
- `generators`
- customization

lib/index.d.ts
```diff
@@ -485,4 +487,29 @@ declare module 'openid-client' {
...
+     */
+    passReqToCallback: boolean
+    /**
+     * The PKCE method to use. When 'true' it will resolve based on the issuer metadata, when 'false' no PKCE will be
+     * used. Default: 'false'
+     */
+    usePKCE: boolean | string
+  }
+
+  export type StrategyVerifyCallbackUserInfo<TUser> = (tokenset: TokenSet, userinfo?: object, done?: (err: any, user?: TUser) => void) => void)
```
theogravity: I had to make `done` optional because typescript will not allow an optional parameter followed by a required one.

Bessonov: If I read [it correctly](https://github.com/panva/node-openid-client/blob/master/docs/README.md#new-strategyoptions-verify) (I'm not familiar with it), then neither `done` nor `userinfo` should be optional in this type. For use case without `userinfo` you have the type `StrategyVerifyCallback` below. Or do I'm missing something?

theogravity: `done` should not be optional, but because `userinfo` is, doing the following is illegal in TS:

`fn (userinfo?, done)`

I looked at the tests for the `Strategy` class in this project, and saw two usages for the verify callback:

`(tokenset, done)
`(tokenset, userinfo, done)`

The docs has `userinfo` as optional, and the tests for the `Strategy` also reflects both usages as well.

Bessonov: But for this reason you have:
```ts
verify: StrategyVerifyCallbackUserInfo<TUser> | StrategyVerifyCallback<TUser>
```

Then it should be enough:
```ts
export type StrategyVerifyCallbackUserInfo<TUser> = (tokenset: TokenSet, userinfo: object, done: (err: any, user?: TUser) => void) => void
export type StrategyVerifyCallback<TUser> = (tokenset: TokenSet, done: (err: any, user?: TUser) => void) => void
```

theogravity:
That should complete the definitions!

I based the error defs on what I saw here:
https://github.com/panva/node-openid-client/blob/master/lib/errors.js

Feel free to update it if it looks incorrect in any way

lib/index.d.ts
theogravity: Oh, I see. Thanks! Will update.

lib/index.d.ts
theogravity: My linter fails if it's not included since it's implemented in `IIssuer`

lib/index.d.ts
```diff
@@ -0,0 +1,644 @@
...
+  import { GotOptions } from 'got'
+
+  type HttpRequestOptions = GotOptions<null>
+  type CustomHttpOptionsProvider = (options: HttpRequestOptions) => HttpRequestOptions
+
+  /**
+   * @see https://github.com/panva/node-openid-client/blob/master/lib/index.js
+   */
+  export const custom : {
+    setHttpOptionsDefaults(params: ISetHttpOptionsDefaults) : void
```
panva: params here is actually following the `GotOptions` interface too

lib/index.d.ts
```diff
@@ -0,0 +1,644 @@
+/// <reference types="@panva/jose" />
```
panva: someone care to explain to a newbie here why this needs to be referenced but `got` doesn't?

theogravity: It should be included. I'll update.

https://stackoverflow.com/questions/48216163/what-does-reference-types-node-mean

lib/index.d.ts
theogravity: ok

theogravity:
The CI needs a kick on one of them - been stalled for hours on npm install

panva:
I’ll retrigger with an empty push, UI doesn’t seem to work all that well ;)

panva:
One more thing - the inline docs in the ts definition - does it show up in intellisence for developers? In my experience it doesn't so i'd rather remove it and only have regular docs to avoid having to maintain two places?

theogravity:
Just realized - I was testing in my own project, but since this is to be included in yours, the module declaration isn't necessary.

I was able to get comments in VS Code after doing that. Try again.

If you don't get the `got` options, try `npm i @types/got --save-dev`

panva:
@ulrichb @YangusKhan @RMHonor @big-kahuna-burger @Bessonov 

I'm looking for more ✅ from you guys, please `npm i theogravity/node-openid-client#typescript-defs` and check if this fits to replace @types/openid-client

lib/index.d.ts
```diff
@@ -638,5 +640,9 @@ export namespace errors {
    */
   export class RPError extends Error {
     response?: IncomingMessage
+    jwt?: string
+    jwt?: string
```
theogravity: you have jwt defined twice

big-kahuna-burger:
> @ulrichb @YangusKhan @RMHonor @big-kahuna-burger @Bessonov
> 
> I'm looking for more ✅ from you guys, please `npm i theogravity/node-openid-client#typescript-defs` and check if this fits to replace @types/openid-client

LGTM, Strategy is in which I wished for. Thanks @theogravity 🚀

lib/index.d.ts
```diff
@@ -0,0 +1,648 @@
...
+
+  /**
+   * Performs an arbitrary grant_type exchange at the token_endpoint.
+   */
+  grant(body: IGrantBody, extras?: IGrantExtras): Promise<TokenSet>
+
+  /**
+   * Introspects a token at the Authorization Server's introspection_endpoint.
+   */
+  introspect(token: string, tokenTypeHint?: string, extras?: IIntrospectExtras): Promise<object>
```
ulrichb: Here a more detailed response type would bee cool.

For example `active: boolean` is a required response property. See  https://tools.ietf.org/html/rfc7662#section-2.2.

Maybe just copy [`IntrospectionResponse`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d8907f303d75056e0710a72bad713179fbc78240/types/openid-client/index.d.ts#L93)?

lib/index.d.ts
```diff
@@ -0,0 +1,648 @@
...
+
+  /**
+   * Fetches the OIDC userinfo response with the provided Access Token. Also handles signed and/or
+   * encrypted userinfo responses. When TokenSet is provided as an argument the userinfo sub property
+   * will also be checked to match the on in the TokenSet's ID Token.
+   *
+   * @param accessToken Access Token value. When TokenSet instance is provided its access_token property
+   * will be used automatically.
+   */
+  userinfo(accessToken: TokenType): Promise<object>
```
ulrichb: Please use `{ readonly [name: string]: unknown }` as return type instad of `object`.

This is still type safe (because of the `unknown` type) and avoids reference errors.

See [this sample](https://www.typescriptlang.org/play/#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzg2RlQHkAjAK3AwAoBKeALnhypoG4BYAKDDwDOOCCAB0EHAHNahYmXaYGogcnL1uPXqEiwEKdNjwEQREhWqYA6lgwALAJKpQADxAwGzeAG9jUYHggAT3gAbVQoAFsQFgEMGCxUSQBdFjQAa1QcAHd8AF8NflQhEXEpGRM5cxprO0cXNyUVNQ0gA) for the difference:

image

ulrichb:
... also unit test (+ execution in the build script) would be cool.

As a starting point you could use these: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/openid-client/openid-client-tests.ts

lib/index.d.ts
panva: better now?

lib/index.d.ts
panva: better now?

panva:
> ... also unit test (+ execution in the build script) would be cool.
> 
> As a starting point you could use these: [DefinitelyTyped/DefinitelyTyped:types/openid-client/openid-client-tests.ts@`master`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/openid-client/openid-client-tests.ts) [](/DefinitelyTyped/DefinitelyTyped/blob/HEAD@{2019-09-06T13:25:23Z}/types/openid-client/openid-client-tests.ts)

@ulrichb can you elaborate on the steps necessary to execute this?

theogravity:
Based on reading the DefinitelyTyped repo:

You write a series of statements, such as:

```
const client = new Client(...)
```

And you write enough of them that it would cover the types that you've written in your definition, then you execute a TS linter against it. If the linter passes, then it's all good.

So what I'm guessing here is if you want to actually have the CI test the defs file, is you write a sample `.ts` file and run the TS linter on it.

theogravity:
Let me know if this is something you'd like me to take the lead on, I'd be happy to set it up, just didn't want to mess with your dependencies if possible.

panva:
I refactored the types, removed a lot of the clutter and added the test to CI (part of the `Lint` job).

Please take a look - @big-kahuna-burger i made changes to the strategy definition, can you please verify it still works as intended for you? I didn't bother with the tests yet.

theogravity:
I would really recommend that a class extends an interface and all methods that take in or return a class, use the interface instead 

from a unit test perspective, if it's a class, then i generally need to be able to create an instance of that class and create spies against it or override its prototype to do any custom implementation

vs if it's an interface, i can generally create a custom class that implements the interface or extends a class that implements it and overuse the methods directly

theogravity:
I would really recommend that anywhere a class is defined as part of a property, parameter, return, etc, that a generic be used. It would make life so much easier when unit testing as I'd be able to create my own custom class and easily override any methods that I need to mock.

Not only that, if I end up writing a custom class that extends it and add new methods, typescript would be able to provide autocompletion support, refactor capabilities, and would recognize my new method because I can specify my custom class as the type for the generic

```
class Issuer <C extends Client> {
     client: C

     constructor(client : C)
}

function blah<C extends Client, I extends Issuer>(client: T, issuer: I) {

}

class CustomIssuer extends Issuer {
   myCustomFn() {

   }
}

function blah<Client, CustomIssuer>(new Client(), new CustomIssuer()) {
   // this wouldn't be valid without generics
   issuer.myCustomFn()
}
```

panva:
@theogravity as-is now i can not only understand the types but first and foremost can continue *maintaining* them. Not using TS myself I don't get the samples you're showing me.

Furthermore the types that the community published to DefinitelyTyped are closer to what the state is now and it seems to be enough.

What do others think?

theogravity:
I'm unsure if your comment was directed towards my comment about introducing back the interface, but I deleted the comment because I realized that maintainability / duplication would be an issue.

a compromise that still works well is using generics instead - what are your thoughts on generics?

panva:
It is directed at your [comment](https://github.com/panva/node-openid-client/pull/184#issuecomment-529130541). "generics" has no meaning to me as non-TS user. If you want to show me please make the update. But no promises, it'll stick :)

theogravity:
Understood. Will do it right now.

panva:
> Understood. Will do it right now.

no rush, i ain't got time today no more.

theogravity:
No problem. Thanks for refactoring it. 

I've added generics where it made sense to. Hopefully the tests demonstrates how powerful it can be. 

I had to add the types package for passport-strategy and got as VS Code was not recognizing the types at all until I added them.

panva:
Thank you all for feedback and contributions. This has landed in 3.7.0

ulrichb:
If the consuming project doesn't have `@types/got` installed, you get:

> error TS7016: Could not find a declaration file for module 'got'.

=> Should we move `@types/got` from the dev-dependencies to the dependencies, so that TS compilation works out of the box?

panva:
I intentionally did not do that, I don't see why non-ts users should have this as a dependency.

Maybe add a typescript readme section with this detail? Is that a big deal? Aren't ts users used to installing types? I assumed you do.

RMHonor:
My first instinct was to put it in peer dependencies, as it then at least notifies the user. But perhaps optional dependencies would be more appropriate? Although it's not an approach I've encountered before

panva:
@RMHonor neither of these two applies

optionalDependencies

> If a dependency can be used, but you would like npm to proceed if it cannot be found or fails to install, then you may put it in the optionalDependencies object.

peerDependencies

> In some cases, you want to express the compatibility of your package with a host tool or library, while not necessarily doing a require of this host.

ulrichb:
@panva The (small) problem exists if you want to use `node-openid-client`, but don't use `got` in the consuming project. Then a) you first get the compile error mentioned above and b) you have to add an unnecessary `@types\got` package dependency in the consuming project.

[Optional dependency](https://npm.github.io/using-pkgs-docs/package-json/types/optionaldependencies.html) looks like a good option :).

panva:
> Optional dependency looks like a good option :).

It isn't, it still installs for everyone. It just that it can optionally fail to install.

panva:
> you have to add an unnecessary @types\got package dependency in the consuming project.

How is it unnecessary? If you're using TS you need those types. If you don't you don't have to install it.

RMHonor:
Something like `optionalPeerDependencies` would be perfect. I'd be inclined to say that it's fine to leave if as it is, and a small brief in the documentation around installing the 3rd party type definitions.

ulrichb:
@panva: 
> How is it unnecessary?

If you don't use `got` directly in the consuming project (only indirectly via `node-openid-client`). Still without having `@types\got` installed you immediately get a compile error in this line:

https://github.com/panva/node-openid-client/blob/45402ede49375ed9bd9660f39d544f4483340bff/types/index.d.ts#L10

=> "error TS7016: Could not find a declaration file for module 'got'"

panva:
In that case, i'm keen to remove the type dependency altogether and make it 

```
{
  [key: string]: unknown;
}
```

and jsdoc add `@see https://github.com/sindresorhus/got/tree/v9.6.0#options`

ulrichb:
Why is a normal dependency so bad? Packages often pull dependencies which aren't used in all cases / for all users.

An example: `express` has a dependency to `cookie` (or `cookie-signature`, or `serve-static`) which obviously isn't used by every `express` user.

ulrichb:
Another argument: Also other packages to it that way.

For example `react-intl` (just an example; it was the first one I found in my project):

https://github.com/formatjs/react-intl/blob/899d17239545717283b1d539b011240428e9948d/package.json#L39-L40

panva:
> Why is a normal dependency so bad? Packages often pull dependencies which aren't used in all cases / for all users.

Packages often do, I choose not to. @types/got has three dependencies which also pack other transitive dependencies.

Showcasing bloated packages that do this is not an argument if i'm honest.

I was about to add the dependency then i checked what it means in terms of other transitive dependencies, so instead if someone could contribute the removal of that import altogether replacing it with the actual object with properties that make sense.

panva:
v3.7.1 fixes this.

ulrichb:
>  I'll move forward with the removal of that import altogether replacing it with the actual object with properties that make sense.

👍

ulrichb:
Just tested it. Compiles fine without `@types\got` installed.

Many thanks @panva!

theogravity:
Thank you everyone!

RMHonor:
Thanks a lot @theogravity, really appreciate these definitions

Issue #220, opened 2020-01-09
Support Self-Issued OpenID Provider


CrowdHailer:
**The problem**

I am trying to use this library for my OAuth services but I would also like my clients to be able to manage their own credentials. They should be able to generate a keypair locally and use that for authentication.
This is part of the OIDC.Core 

https://openid.net/specs/openid-connect-core-1_0.html#SelfIssued

**Describe the solution you'd like**
I would like someway to be able to identify that my client is working as self issued provider.

When a issuer is identified as self-issued the acceptable yes should be taken from the token that is passed to the callback. (under the `sub_jwk` claim). Instead of using the value of `jwks_uri`

**A hacked solution**

Having previously set up a client and a issue as standard I can patch the keystore on the issuer to get the desired effect.

```js
const params = client.callbackParams(req);
  client
    .oauthCallback("https://client.example.com/callback", params, {}) // => Promise
    .then(function(tokenSet) {
      const claims = tokenSet.claims();
      issuer.keystore = function functionName() {
        return jose.JWKS.asKeyStore(
          { keys: [tokenSet.claims().sub_jwk] },
          { ignoreErrors: true }
        );
      };

      client
        .callback("https://client.example.com/callback", params, {})
        .then(function(tokenSet) {
          console.log("validated ID Token claims %j", tokenSet.claims());
        });
    });
```

There needs to be a way to configure the application to use the keys in this way.

My suggestion would be to configure `jwks_uri` to look for a `sub_jwk` claim instead of fetching keys

```js
{
  "jwks_uri": "sub_jwk",
  "jwks_uri": "https://self-issued.me"
} 
```

I think either of those would work. Depends on if you want the value to be a valid URL even though it will not be used as a url.

**Note** I would suggest configuring it this way rather than off the issuer. because for development it is useful to be able to have issuers which are tied to localhost.

---

 - [x] i have searched the issues tracker on github for similar requests and couldn't find anything related.

panva:
There's more to it than just using the embedded JWK, i had [this](https://github.com/panva/node-openid-client/compare/feat-siop) stashed in my local repo, see if it works for you. All it takes is for the Issuer instance issuer identifier to be the one from the spec `https://self-issued.me`

CrowdHailer:
Ahh yes I see the comparing the subject with the key fingerprint is a crucial thing to do as well. 

That looks like it would work for me, I'll double check tomorrow.
Are their other considerations that stopped you adding it to the library

[...]

PR #221, opened 2020-01-10
feat: detect self-issued OP and validate ID Token accordingly

CrowdHailer:
Add support for section 7 of openid.core "Self-Issued OpenID Provider"
https://openid.net/specs/openid-connect-core-1_0.html#SelfIssued

- [x] handle special issuer `https://self-issued.me`
- [x] Add tests
- [ ] Add documentation to the README


CrowdHailer:
Opened to handle #220

Issue #231, opened 2020-02-21
JWT Authorized Presenter (azp) validation is too strict


svvac:
Currently, JWT validation throws if the authorized presenter (`azp` field) is not equal to the `client_id` ([see here](https://github.com/panva/node-openid-client/blob/master/lib/client.js#L906)).

However, according to [Google's OIDC docs](https://developers.google.com/identity/protocols/OpenIDConnect#obtainuserinfo), it is specifically different from the `aud` (Audience) field if present (emphasis mine) :

> **`azp`** : The `client_id` of the authorized presenter. This claim is only needed when **the party requesting the ID token is not the same as the audience of the ID token**. This may be the case at Google for hybrid apps where a web application and Android app have a different OAuth 2.0 client_id but share the same Google APIs project.

More specifically, it happens in Android mobile app authentication flow : 

1. The app requests an ID token from Google using the app's cert and client ID, and the target server's client_id
2. Google issues an ID token with `aud` set to the server's client_id and `azp` to the app's client ID
3. The app gets the token and sends it to the server for logging in
4. The server receives the id token, validates it, and issues credentials based on the claims of the id token

AFAICT, this auth flow is currently not possible with node-openid-client

panva:
These validations aren’t strict, thats how the spec is written. What would you propose gets loosened?

svvac:
After reading relevant parts of the spec, it appears that the check in question is indeed mandated : 

> **2 — ID token** : `azp` : OPTIONAL. Authorized party - the party to which the ID Token was issued. If present, it MUST contain the OAuth 2.0 Client ID of this party. This Claim is only needed when the ID Token has a single audience value and that audience is different than the authorized party. It MAY be included even when the authorized party is the same as the sole audience. The `azp` value is a case sensitive string containing a StringOrURI value. 

> **3.1.3.7 — ID token validation** : 5. If an `azp` (authorized party) Claim is present, the Client SHOULD verify that its `client_id` is the Claim Value. 

However, this raises the question : if multiple audiences can be provided, how are the others (the ones that are not `azp`) supposed to validate the ID token if the assertion `aud === azp === client_id` must hold?

The spec indeed considers cases where this is not the case, for instance:

[...]

PR #233, opened 2020-02-27
Add support for extra Client options

svvac:
see #231


svvac:
Supplanted by #234. Sorry for the noise.

PR #234, opened 2020-02-27
feat: support additional authorized parties

### LINKED ISSUE #231: JWT Authorized Presenter (azp) validation is too strict

svvac:
I have a variant ready without `allowedThirdParties: true` support if it's a no-no.

Includes #233 which I'm going to close (what was I thinking splitting this?)

Closes #231

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -334,13 +334,16 @@ export interface IntrospectionResponse {
   [key: string]: unknown;
 }
 
+export interface ClientOptions {
+}
+
 /**
  * Encapsulates a dynamically registered, discovered or instantiated OpenID Connect Client (Client),
  * Relying Party (RP), and its metadata, its instances hold the methods for getting an authorization URL,
  * consuming callbacks, triggering token endpoint grants, revoking and introspecting tokens.
  */
 export class Client {
-  constructor(metadata: ClientMetadata, jwks?: JSONWebKeySet);
+  constructor(metadata: ClientMetadata, jwks?: JSONWebKeySet, options?: ClientOptions);
   [custom.http_options]: CustomHttpOptionsProvider;
   [custom.clock_tolerance]: number;
   metadata: ClientMetadata;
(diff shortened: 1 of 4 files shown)


panva:
> I have a variant ready without `allowedThirdParties: true` support if it's a no-no.

It is a no-no as previously discussed.

svvac:
That's sad, but I can live with it.

panva:
> That's sad, but I can live with it.

Okay.

I've pushed a commit to your upstream, please see if i didn't mess up anything.

panva:
May I also ask you to also rename authorizedThirdParties to additionalAuthorizedParties?

svvac:
LGTM

- [x] Renamed the option to `additionalAuthorizedParties`
- [x] Squashed reviews in appropriate commits

types/index.d.ts
```diff
@@ -334,13 +334,17 @@ export interface IntrospectionResponse {
   [key: string]: unknown;
 }
 
+export interface ClientOptions {
+  additionalAuthorizedParties?: false | string | string[];
```
panva: Remove the false here. It somehow got back in there.

PR #284, opened 2020-09-01
fix(typescript): add missing types

jaulz:
Some helpful properties were missing in the types.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -347,7 +347,7 @@ export class Client {
   [custom.http_options]: CustomHttpOptionsProvider;
   [custom.clock_tolerance]: number;
   metadata: ClientMetadata;
-  issuer: Issuer;
+  issuer: Issuer<this>;
 
   /**
    * Returns the target authorization redirect URI to redirect End-Users to using the provided parameters.
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -69,7 +69,8 @@ export interface ClientMetadata {
   authorization_encrypted_response_alg?: string;
   authorization_encrypted_response_enc?: string;
   authorization_signed_response_alg?: string;
-
+  scopes_supported?: string[];
+  
   [key: string]: unknown;
 }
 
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -69,7 +69,8 @@ export interface ClientMetadata {
   authorization_encrypted_response_alg?: string;
   authorization_encrypted_response_enc?: string;
   authorization_signed_response_alg?: string;
-
+  scopes_supported?: string[];
+
```
panva, line 73: `scopes_supported` is not an [IANA registered client metadata property](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#client-metadata). Please remove this blob.

`scopes_supported` could potentially go to the `IssuerMetadata` interface, but as the library never reacts to its presence it's not included. I'm keen to only include properties that change the client/issuer behaviour. So it's either all or nothing and i think the inclusion of `[key: string]: unknown;` handles all possible metadata for anyone.

types/index.d.ts
```diff
@@ -346,6 +347,7 @@ export class Client {
   [custom.http_options]: CustomHttpOptionsProvider;
   [custom.clock_tolerance]: number;
   metadata: ClientMetadata;
+  issuer: Issuer<this>;
```
panva, line 350: there's also a [`static` version of it](https://github.com/panva/node-openid-client/blob/v3.15.9/lib/client.js#L1565-L1579), please add that too.

panva (changes requested):

types/index.d.ts
jaulz: Done.

jaulz:
Thanks for the detailed review!

types/index.d.ts
```diff
@@ -346,6 +346,8 @@ export class Client {
...
+  static issuer: Issuer<Client>;
```
panva, line 350:
```suggestion
  static issuer: Issuer<TClient extends Client>;
```

types/index.d.ts
```diff
@@ -346,6 +346,8 @@ export class Client {
   [custom.http_options]: CustomHttpOptionsProvider;
   [custom.clock_tolerance]: number;
   metadata: ClientMetadata;
+  issuer: Issuer<this>;
+  static issuer: Issuer<TClient extends Client>;
```
panva, line 350:
```suggestion
  static issuer: Issuer<Client>;
```

panva:
@jaulz thanks for the contribution!

PR #289, opened 2020-09-11
docs: fix HTTP debugging example

ctavan:
The parameter structure has changed due to the got major version upgrade
which was not yet reflected in the documented example.


panva:
Thank you!

PR #290, opened 2020-09-11
docs: PKCE is used by default in passport strategy now

tronghn:
Changed in https://github.com/panva/node-openid-client/commit/56f9fe7171ccc1bec6427d4f9bc45e419150ab4d


panva:
thank you!

PR #292, opened 2020-09-19
Replace leftover deprecated "verb" with "method"

madarche:
This PR has only 1 true fix: in the doc of https://github.com/panva/node-openid-client/blob/master/docs/README.md#clientuserinfoaccesstoken-options

The other substitutions have been done for consistency. The substitution in the error message is debatable, but code should not rely on error messages, error messages are not an API. I can delete this change if you want though 🙂


panva (approved):

panva:
Thank you!

PR #295, opened 2020-10-02
feat: add callback extras to strategy options

jonalu:
### The problem
Currently the Strategy does not support client assertions making it troublesome to use with Passport.

### Proposed solution
Pass option `extras: CallbackExtras` to the `StrategyOptions` interface like so:

```
const strategy = new Strategy({
        client,
        params: {
            scope: 'openid profile',
        },
        extras: {
            clientAssertionPayload: {
                aud: metadata.issuer
            }
        }
    }, (tokenSet: TokenSet, done: (_err: any, _user?: any) => void) => {
        done(null, tokenSet)
    });
```

The function `OpenIDConnectStrategy.prototype.authenticate` is changed to include the `client._extras` as the last parameter to `client.callback`

### DIFF
--- a/lib/passport_strategy.js
+++ b/lib/passport_strategy.js
@@ -30,6 +30,7 @@ function OpenIDConnectStrategy({
   passReqToCallback = false,
   sessionKey,
   usePKCE,
+  extras = {},
 } = {}, verify) {
   if (!(client instanceof BaseClient)) {
     throw new TypeError('client must be an instance of openid-client Client');
@@ -50,6 +51,7 @@ function OpenIDConnectStrategy({
   this._usePKCE = usePKCE;
   this._key = sessionKey || `oidc:${url.parse(this._issuer.issuer).hostname}`;
   this._params = cloneDeep(params);
+  this._extras = cloneDeep(extras);
 
   if (!this._params.response_type) this._params.response_type = resolveResponseType.call(client);
   if (!this._params.redirect_uri) this._params.redirect_uri = resolveRedirectUri.call(client);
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -38,3 +38,6 @@ package-lock.json
 
 # TS test
 types/*.js
+
+# IDEA
+.idea
```
panva, line 43: remove this change

lib/passport_strategy.js
```diff
@@ -82,6 +84,7 @@ OpenIDConnectStrategy.prototype.authenticate = function authenticate(req, option
       throw new TypeError('authentication requires session support');
     }
     const reqParams = client.callbackParams(req);
+
```
panva, line 87: remove this change

test/passport/passport_strategy.test.js
```diff
@@ -486,7 +520,7 @@ describe('OpenIDConnectStrategy', () => {
     });
 
     it('throws when userinfo is requested but no access_token was returned', function (next) {
-      const strategy = new Strategy({ client: this.client }, (tokenset, userinfo, done) => {}); // eslint-disable-line
+      const strategy = new Strategy({ client: this.client }, (_tokenset, _userinfo, _done) => {}); // eslint-disable-line
```
panva, line 523: remove this change

types/index.d.ts
```diff
@@ -675,6 +675,11 @@ export interface StrategyOptions<TClient extends Client> {
    * Authorization Request parameters. The strategy will use these.
    */
   params?: AuthorizationParameters;
+
+  /**
+   * Extra Authorization Request parameters.
+   */
+  extras ?: CallbackExtras;
```
panva, line 682:
```suggestion
  /**
   * "extras" argument value passed to the client.callback() call.
   */
  extras?: CallbackExtras;
```

panva (changes requested):
There's a lot that seems to have been done by whatever IDEs automated formatting. Please remove those change and only keep what's new.

jonalu:
> There's a lot that seems to have been done by whatever IDEs automated formatting. Please remove those change and only keep what's new.

@panva Thanks! Do you mind having another look at it?

panva (approved):

PR #324, opened 2021-01-18
feat(typescript): add userinfo response generics

filipesmedeiros:
By using generics, we can enforce proper type checking without the need of casting. This is just the first, because `[key: string]: unknown` is used some more in the type declarations.

### DIFF
--- a/types/openid-client-tests.ts
+++ b/types/openid-client-tests.ts
@@ -113,6 +113,21 @@ async (req: IncomingMessage) => {
 
     //
 
+    const userinfoGeneric = await client.userinfo<{ someProp: number }>(callbackResponse);
+    console.log(userinfoGeneric.someProp);
+
+    //
+
+    const userinfoAddressGeneric = await client.userinfo<{ someProp: number }, { street: string }>(callbackResponse);
+    console.log(userinfoAddressGeneric.address ? userinfoAddressGeneric.address.street : undefined);
+
+    //
+
+    const userinfoOverride = await client.userinfo<{ email_verified: string }>(callbackResponse);
+    console.log(userinfoOverride.email_verified);
+
+    //
+
     await client.requestResource('https://rs.example.com/resource', 'access token', { headers: { Accept: 'application/json' } });
     const resource = await client.requestResource('https://rs.example.com/resource', callbackResponse, { headers: { Accept: 'application/json' } });
     console.log(resource.body.byteLength);
--- a/types/openid-client-tests.ts
+++ b/types/openid-client-tests.ts
@@ -119,12 +119,12 @@ async (req: IncomingMessage) => {
     //
 
     const userinfoAddressGeneric = await client.userinfo<{ someProp: number }, { street: string }>(callbackResponse);
-    console.log(userinfoAddressGeneric.address ? userinfoAddressGeneric.address.street : undefined);
+    console.log(userinfoAddressGeneric.address ? userinfoAddressGeneric.address.street.substring(0) : undefined);
 
     //
 
     const userinfoOverride = await client.userinfo<{ email_verified: string }>(callbackResponse);
-    console.log(userinfoOverride.email_verified);
+    console.log(userinfoOverride.email_verified.substring(0));
 
     //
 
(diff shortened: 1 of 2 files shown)


panva:
Please make sure the PR contains no unnecessary changes such as changes to just padding, extra spaces, or quotation marks. Else this cannot be reviewed. Thank you.

filipesmedeiros:
Hey! Yeah I tried to be careful but I guess it slipped though anyway ahah, gonna fix it!

filipesmedeiros:
If you could check again, that would be awesome, thanks

filipesmedeiros:
I can't figure out why https://github.com/panva/node-openid-client/blob/main/types/openid-client-tests.ts#L112 uses the string `'some user info name'` there, but it's making the generics not work. What do you think @panva ?

panva:
> I can't figure out why [https://github.com/panva/node-openid-client/blob/main/types/openid-client-tests.ts#L112](https://github.com/panva/node-openid-client/blob/main/types/openid-client-tests.ts?rgh-link-date=2021-01-18T16%3A03%3A17Z#L112) uses the string `'some user info name'` there, but it's making the generics not work. What do you think @panva ?

because of `[key: string]: unknown;` on the current types. As i said in my earlier comment, your PR must be backwards compatible so the defaults you're assigning probably need to be `[key: string]: unknown;` too.

filipesmedeiros:
I understand, and did that. But imo (of course you can take it as you want) this TS change would be worth a major change. Generics should replace having `unknown`s. Of course, if you don't think so, we can just keep adding generics while keeping them there!

panva:
> I understand, and did that. But imo (of course you can take it as you want) this TS change would be worth a major change. Generics should replace having `unknown`s. Of course, if you don't think so, we can just keep adding generics while keeping them there!

I'm not on board with a new major just because of types. I think you should be able to keep the current behaviour and remove unknowns when one uses generics by having your `TUserInfo` default be `{ [key: string]: unknown; }` object.

filipesmedeiros:
> I think you should be able to keep the current behaviour and remove unknowns when one uses generics by having your TUserInfo default be { [key: string]: unknown; } object.

And then, to remove the unknowns, we just have to pass `{}` or something to the generic, right? That sounds great actually!

panva:
I think you're good to add tests of the generics now.

filipesmedeiros:
Hey @panva can you see if this is something you like?

types/openid-client-tests.ts
```diff
@@ -113,6 +113,21 @@ async (req: IncomingMessage) => {
 
     //
 
+    const userinfoGeneric = await client.userinfo<{ someProp: number }>(callbackResponse);
+    console.log(userinfoGeneric.someProp);
```
panva, line 117:
```suggestion
    console.log(userinfoGeneric.someProp.toFixed(1));
```

types/openid-client-tests.ts
```diff
@@ -113,6 +113,21 @@ async (req: IncomingMessage) => {
...
+
+    //
+
+    const userinfoAddressGeneric = await client.userinfo<{ someProp: number }, { street: string }>(callbackResponse);
+    console.log(userinfoAddressGeneric.address ? userinfoAddressGeneric.address.street : undefined);
```
panva, line 122:
```suggestion
    console.log(userinfoAddressGeneric.address ? userinfoAddressGeneric.address.street.substring(0) : undefined);
```

types/openid-client-tests.ts
```diff
@@ -113,6 +113,21 @@ async (req: IncomingMessage) => {
...
+
+    //
+
+    const userinfoOverride = await client.userinfo<{ email_verified: string }>(callbackResponse);
+    console.log(userinfoOverride.email_verified);
```
panva, line 127:
```suggestion
    console.log(userinfoOverride.email_verified.substring(0));
```

panva (changes requested):

panva:
@filipesmedeiros thank you, keep 'em coming!

filipesmedeiros:
Nice! Thanks for the awesome package, do you have a prediction for when this can go live? Eheh
I'll try to finish the other generics when I have the time!

panva:
> do you have a prediction for when this can go live?

Unless a fix is needed in the meantime I will release a new minor version over the weekend.

PR #328, opened 2021-01-29
feat: allow options.https.pfx for mTSL

cplussharp:
so mTLS can be used without options.https.key and options.https.certificate

see #326


panva (approved):

panva:
Thank you @cplussharp!

cplussharp:
That was fast, thanks for merging.

Issue #281, opened 2020-08-04
error while trying to connect several times


JennyMet:
HI,

I use the following code which works, However after few success calls (3-10), I sometimes get internal server error with the following  error,

I use the `3.15.9` 
Any idea what could be wrong? 

```
Error: did not find expected authorization request details in session, req.session["oidc:accounts.rvm.com"] is undefined
at /opt/node_app/app/node_modules/openid-client/lib/passport_strategy.js:125:13
at OpenIDConnectStrategy.authenticate (/opt/node_app/app/node_modules/openid-client/lib/passport_strategy.js:173:5)
at attempt (/opt/node_app/app/node_modules/passport/lib/middleware/authenticate.js:366:16)
at authenticate (/opt/node_app/app/node_modules/passport/lib/middleware/authenticate.js:367:7)
at /opt/node_app/app/src/logon.js:92:7
at Layer.handle [as handle_request] (/opt/node_app/app/node_modules/express/lib/router/layer.js:95:5)
at next (/opt/node_app/app/node_modules/express/lib/router/route.js:137:13)
at Route.dispatch (/opt/node_app/app/node_modules/express/lib/router/route.js:112:3)
at Layer.handle [as handle_request] (/opt/node_app/app/node_modules/express/lib/router/layer.js:95:5)
at /opt/node_app/app/node_modules/express/lib/router/index.js:281:22
```

My code from the stack is
```
at /opt/node_app/app/src/logon.js:92:7
```
which is the end of the code

```
    })(req, res, next);   //here is the error

```

This is the full code (I pass the `app` which is simply express server) , am I doing something wrong? 

This is the **index.js**

```
const express = require('express');
const logon = require('./logon');

const app = express();
const port = process.env.PORT || 4000;

logon(app)
  .then(() => {
    console.log('process started');
  });
app.use(express.json());

app.listen(port,
  () => console.log(`listening on port: ${port}`));
```
This is the l**ogon.js**

```
const { Issuer, Strategy } = require('openid-client');
const cookieParser = require('cookie-parser');
const cookieSession = require('cookie-session');
const azpi = require('./azpi');
const bodyParser = require('body-parser');
const passport = require('passport');


module.exports = async (app) => {

  let oSrv;
  const durl = `${process.env.srvurl}/.well-known/openid-configuration`;
  try {
    oSrv = await Issuer.discover(durl);
  } catch (err) {
    console.log('error occured', err);
    return;
  }

[...]

Issue #237, opened 2020-03-02
Did not find expected authorization request details in session.


abhinavtcs:
**Describe the bug**
Did not find expected authorization request details in session.

```
Error: did not find expected authorization request details in session, req.session["oidc:federate.xh1.lilly.com"] is undefined
    at C:\Users\workspace\Auth Module\scout_auth_module\node_modules\openid-client\lib\passport_strategy.js:126:13
    at OpenIDConnectStrategy.authenticate (C:\Users\C286160\workspace\Auth Module\scout_auth_module\node_modules\openid-client\lib\passport_strategy.js:174:5)
    at attempt (C:\Users\workspace\Auth Module\scout_auth_module\node_modules\Passport\lib\middleware\authenticate.js:366:16)
    at authenticate (C:\Users\workspace\Auth Module\scout_auth_module\node_modules\Passport\lib\middleware\authenticate.js:367:7)
    at router.use (C:\Users\workspace\Auth Module\scout_auth_module\index.js:98:8)
    at Layer.handle [as handle_request] (C:\Users\workspace\Auth Module\testApp\node_modules\express\lib\router\layer.js:95:5)
    at trim_prefix (C:\Users\workspace\Auth Module\testApp\node_modules\express\lib\router\index.js:317:13)
    at C:\Users\workspace\Auth Module\testApp\node_modules\express\lib\router\index.js:284:7
    at Function.process_params (C:\Users\workspace\Auth Module\testApp\node_modules\express\lib\router\index.js:335:12)
    at next (C:\Users\workspace\Auth Module\testApp\node_modules\express\lib\router\index.js:275:10)
```

**To Reproduce**
```
exports.authenticate = async (router) => {
    router.set('trust proxy', true);
    router.use(session({
        saveUninitialized:'true',
        resave:'true',
        secret: "hello",
        name: 'id',
        cookie: {
          httpOnly: 'false',
          maxAge: 1800000,
          path: '/',
          sameSite: undefined,
          secure: undefined,
        },
        proxy: undefined,
        rolling:'true',
        unset: 'keep',
      }))

    router.use(bodyParser.json());
    router.use(bodyParser.urlencoded({ extended: true }));

    router.use(passport.initialize());
    router.use(passport.session());

    router.use('/login',
    (req, res, next) => {
      const redirectUri = `http://localhost:3000/auth/callback`;
      const state = req.originalUrl;

[...]

PR #335, opened 2021-02-17
fix(passport-strategy): ensure session is saved before redirecting

josepapaianni:
Some session stores could be a little bit slow sometimes, specially the aaS ones during the initial request. 

Also looks like it's the [right](https://github.com/expressjs/session#sessionsavecallback) way to handle sessions while doing redirects.

Probably will solve issues like: #237 #281

### DIFF
--- a/lib/passport_strategy.js
+++ b/lib/passport_strategy.js
@@ -116,6 +116,13 @@ OpenIDConnectStrategy.prototype.authenticate = function authenticate(req, option
         }
       }
 
+      await new Promise((resolve, reject) => {
+        req.session.save((err) => {
+          if (err) reject(err);
+          resolve();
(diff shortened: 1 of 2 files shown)


josepapaianni:
*Fixing compatibility with Node 10 before reopening

panva:
@josepapaianni thank you for attempting here but ultimately I would not accept even a passing PR here. The responsibility for ensuring sessions get saved lies with passport or the session mechanism itself, not with a passport strategy that is agnostic to the session store module or its "store" one uses.

josepapaianni:
Got it, thanks for the quick response. Do you mind to add a warning about this in the Strategy docs? (As a well known issue)

PR #351, opened 2021-04-14
Allow HEAD in callbackParams

donavon:
http `HEAD` should perform the same actions as `GET`, but not return a body. This fix will return the same headers.

Chrome (especially on Windows) is seen hitting our server logs against `HEAD /login` and this is causing an "invalid IncomingMessage method" error. This PR fixes the problem.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -338,6 +338,7 @@ module.exports = (issuer, aadIssValidation = false) => class Client extends Base
     if (isIncomingMessage) {
       switch (input.method) {
         case 'GET':
+        case 'HEAD':
           return pickCb(url.parse(input.url, true).query);
         case 'POST':
           if (input.body === undefined) {


panva:
Hi @donavon 

I think you should look into why you're receiving HEAD requests on callback paths in the first place. Any browser redirects will end up with a GET, `form_post` will end up with a POST. Those are handled.

donavon:
It doesn't matter why. This is a bug.

> The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response. The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request.

[RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4)

On Chrome/Windows a browser redirect is doing a HEAD. I don't know why, but it is.

panva:
@donavon would you mind sharing a screenshot of the network tab with the AS redirect?

panva:
Please let me know if you have found out the why.

PR #357, opened 2021-04-22
feat: add abort control over Device Flow Handle polling

panva:
Example using AbortController.

```js
const ac = new AbortController()

setTimeout(() => ac.abort(), 10000)

try {
  await handle.poll({ signal: ac.signal })
} catch (err) {
  console.log(err)
}
```

Example using handle.abort().

```js
setTimeout(() => handle.abort(), 10000)

try {
  await handle.poll({ signal: ac.signal })
} catch (err) {
  console.log(err)
}
```

resolves #355
closes #356

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -793,7 +793,8 @@ first place.
 
 <!-- TOC DeviceFlowHandle START -->
 - [Class: &lt;DeviceFlowHandle&gt;](#class-deviceflowhandle)
-  - [handle.poll()](#handlepoll)
+  - [handle.poll([options])](#handlepolloptions)
+  - [handle.abort()](#handleabort)
   - [handle.user_code](#handleuser_code)
   - [handle.verification_uri](#handleverification_uri)
   - [handle.verification_uri_complete](#handleverification_uri_complete)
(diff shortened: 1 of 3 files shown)


panva:
@NotMyself please have a look.

I went with a standard AbortController approach thats making rounds in Node.js core, as well as an instance method to abort for runtimes below Node.js 15 where AbortController was introduced.

The important thing is to make the original promise resolve or reject, it should not be left in a pending state. I went with a rejection when aborted.

NotMyself:
That is much cleaner than what I was doing. If I am reading the sample correctly I should be doing the following in my extension code. Is that correct?

```typescript
async (progress, token) => {
    vscode.commands.executeCommand(
      'vscode.open',
      vscode.Uri.parse(handle.verification_uri_complete)
    );
    
    const ac = new AbortController();
    token.onCancellationRequested(() => {
      ac.abort();
    });

    const tokenSet = await handle.poll({ signal: ac.signal });

    await keytar.setPassword(
      SECRET_KEY_SERVICE_NAME,
      'token_set',
      JSON.stringify(tokenSet)
    );

    authStatusEventEmitter.fire(tokenSet);
  }

```

panva:
Yes, I think so. And it will reject handle.poll() with the next interval check. Are you okay with me proceeding with this PR then?

NotMyself:
Yeah, this works for me. Thank you for such a wonderful library. So easy to use.

panva:
> Yeah, this works for me. Thank you for such a wonderful library. So easy to use.

Please consider [supporting](https://github.com/panva/node-openid-client?sponsor=1) the library if it provides value to you or your company and this was of help to you. Supporting the library means, amongst other things, that such support and new features will be available in the future.

NotMyself:
@panva I work at Auth0 and we do sponsor you. 😁

panva:
> @panva I work at Auth0 and we do sponsor you. 😁

standard reply ;)

PR #371, opened 2021-05-31
Make static Issuer methods generic

mikaelkaron:
Allows the client returned from either `Issuer<TClient>.discover` or `Issuer<TClient>.webfinger` to be generically typed.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -676,13 +676,13 @@ export class Issuer<TClient extends Client> {
    * performs both openid-configuration and oauth-authorization-server requests.
    * @param issuer Issuer Identifier or metadata URL
    */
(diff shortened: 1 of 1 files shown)


panva (changes requested):
Hi @mikaelkaron,

please make this a non-breaking change, e.g. by defaulting to the before used values and add tests verifying the updated functionality to `types/openid-client-tests.ts`

mikaelkaron:
Not really sure how I'd add tests for this type change. But it's now a non-breaking change.

panva (changes requested):
Just missing tests in `types/openid-client-tests.ts` that would confirm the functionality and ensure future type changes don't break this addition.

panva:
Ping me here if you'd like this reopen after you've added the changes requested.

PR #407, opened 2021-09-20
Populate 'ath' claim in DPoP when access token is used

### LINKED ISSUE #406: update DPoP to an updated draft
**Describe the bug**
I attempted to create a simple client/server example of DPoP usage, using node-openid-client and node-oidc-provider.  The client logs in using DPoP, then attempts to fetch `userinfo`.  The `userinfo` request fails with `invalid_token (invalid DPoP key binding)`.  Looking into the server side more closely, it appears that the `userinfo` request does have a DPoP signature, but it is missing the `ath` field.  According to draft-03 of DPoP, it seems like this field is required "[w]hen the DPoP proof is used in conjunction with the presentation of an access token", so the server's interpretation is correct here.

So I think what is needed here is to extend the [call to `dpopProof()` in the `request()` method](https://github.com/panva/node-openid-client/blob/992fec49e2908323547a3e9010613f74b2e88ad4/lib/helpers/request.js#L33) so that it populates the `ath` parameter in the DPoP payload when the request is made with an access token.

**To Reproduce**

Provider and client JS code provided in [this gist](https://gist.github.com/bifurcation/dc676adf3c5948ad2bb2e0c24029c785#file-server-js).  The scripts depend on a few modules, and assume that the domain names `oidc-client.invalid` and `oidc-provider.invalid` are mapped to localhost in `/etc/hosts` or equivalent.

Steps to reproduce the behaviour:
1.  In one window: `node server.js`
2.  In another window: `NODE_TLS_REJECT_UNAUTHORIZED=0 node client.js`
3.  Observe error reported in client window.

**Expected behaviour**
The `client.js` script prints a set of claims returned by the provider.

**Environment:**
 - openid-client version: 4.8.0
 - oidc-provider version: 7.8.0
 - node version: v14.17.0

**Additional context**

 - [x] the bug is happening on latest openid-client too.
 - [x] i have searched the issues tracker on github for similar issues and couldn't find anything related.

bifurcation:
Fixes #406 

According to draft-03 of DPoP:

```
   When the DPoP proof is used in conjunction with the presentation of
   an access token, see Section 7, the DPoP proof also contains the
   following claim:

   *  "ath": hash of the access token (REQUIRED).  The value MUST be the
      result of a base64url encoding (with no padding) the SHA-256 hash
      of the ASCII encoding of the associated access token's value.
```

The DPoP implementation in this repository currently does not provide the `ath` claim when used with an access token (e.g., in `userinfo()` and `requestResource()`).  This PR updates the DPoP proof construction code so that if an access token is provided, then the `ath` parameter is included in the proof.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ openid-client.
   - RP-Initiated Logout
 - [Financial-grade API - Part 2: Read and Write API Security Profile (FAPI) - ID2][feature-fapi]
 - [JWT Secured Authorization Response Mode for OAuth 2.0 (JARM) - ID1][feature-jarm]
-- [OAuth 2.0 Demonstration of Proof-of-Possession at the Application Layer (DPoP) - draft 01][feature-dpop]
+- [OAuth 2.0 Demonstration of Proof-of-Possession at the Application Layer (DPoP) - draft 03][feature-dpop]
 
 Updates to draft specifications (DPoP, JARM, and FAPI) are released as MINOR library versions,
 if you utilize these specification implementations consider using the tilde `~` operator in your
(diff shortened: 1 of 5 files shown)


bifurcation:
I have also verified that this branch interoperates with node-oidc-provider, using the scripts linked in #406.

panva:
@bifurcation thank you for the PR, i've made a few small adjustments and will release a new minor (this is a feat, not a fix).

Issue #447, opened 2021-12-20
Failure on node 17.x.x because of engines spec in package.json


lucasgonze:
**Describe the bug**

Use openid-client@5.1.0 in a context with Node 17.x.x and engines-strict. You'll get this error:

`The engine "node" is incompatible with this module. Expected version "^12.19.0 || ^14.15.0 || ^16.13.0". Got "17.1.0"`

(I note that getting an error instead of a warning means that engines-strict is being set in the CI, probably. If you decide that 17.x.x is incompatible or not worth the trouble, that's your call).

**To Reproduce**

nvm use 17.0.1

**Environment:**
 - openid-client version: 5.1.0
 - node version: 17.1.0

**Additional context**

You can see the error in a real-world CI here: https://app.circleci.com/pipelines/github/facebookincubator/fbc-js-core/445/workflows/6544e2ac-52fe-447b-a068-b23c8e88b8a6/jobs/3357

![image](https://user-images.githubusercontent.com/111581/146828000-bace2a0e-d597-46d6-9c0c-fcd1b5276dc8.png)

 - [ ] the bug is happening on latest openid-client too.
 - [ ] i have searched the issues tracker on github for similar issues and couldn't find anything related.

panva:
Please see v5.x release notes. This is in place very much on purpose as I only intend to support Node.js LTS releases.

lucasgonze:
> Major Node.js versions enter Current release status for six months, which gives library authors time to add support for them. After six months, odd-numbered releases (9, 11, etc.) become unsupported, and even-numbered releases (10, 12, etc.) move to Active LTS status and are ready for general use.

Ok, understood. I'll upgrade to v18 ;)

panva:
v18 goes LTS in October 2022, that's when v18 will be added to engines.

lucasgonze:
It's not your intention to disallow non-LTS versions, only to limit version-specific fixes to LTS versions, right?

There is a conflict with yarn's policy of strict mode: https://github.com/yarnpkg/rfcs/pull/69 . That is, openid-client will always fail for an odd-numbered version of Node when installed using Yarn (but not npm).

lucasgonze:
I have tested a workaround: setting engineStrict to false. This worked. If this is in sync with your intentions, here is a PR:

https://github.com/panva/node-openid-client/pull/448

panva:
If you want yarn to install despite the lack of support for the engine, use yarn's --ignore-engines.

[...]

PR #448, opened 2021-12-20
Set engineStrict false in order to enable Yarn to install

lucasgonze:
Set engineStrict to false in order to enable Yarn to install when an odd-numbered version of Node is in use.

Signed-off-by: Lucas Gonze <lucas@gonze.com>


panva:
https://github.com/panva/node-openid-client/issues/447#issuecomment-998263129

Issue #466, opened 2022-02-03
OIDC JWK endpoint expects application/jwk-set+json but only gets accept application/json (406 Not Acceptable)


frederik:
**Describe the bug**

The jwk endpoint to retrieve the JWK sets seems to usually accept _application/json_ as an accept header.
However, some OIDC provider implementations explicitly expect Accept: 'application/jwk-set+json' (which is [valid ](https://datatracker.ietf.org/doc/html/rfc7517) for JWKS according to the spec). As a result they send back a _406 Not Acceptable_.

Other libraries seem to have added this header, see [Tomcat OIDC auth](https://github.com/boylesoftware/tomcat-oidcauth/issues/19) and others.

I do not have access to a public endpoint with this behavior, but  would be able to verify once the header is added.

**Expected behaviour**

Setting the headers to Accept: 'application/json, application/jwk-set+json'
results in a successful request.

The line responsible is the GET request in function getKeyStore (lib/helpers/issuer.js). The following headers result in a successful request

```
headers: {
    Accept: 'application/json, application/jwk-set+json',
}
```

I'd be happy to submit a pull request, if this is accepted as a bug.

**Environment:**
 - openid-client version: 5.1.1
 - node version: 16.13.1

**Additional context**

 - [x] the bug is happening on latest openid-client too.
 - [x] i have searched the issues tracker on github for similar issues and couldn't find anything related.

Thanks for this great library and the work you put into it 😃 !

panva:
Feel free to open a PR, I'll have a look at it!

PR #467, opened 2022-02-03
#466 adding application/jwk-set+json for JWTS calls

frederik:
Adding application/jwk-set+json as an accepted header, since some OIDC providers are stricter and will return a 406 error code otherwise.


panva (dismissed):
Please update `test/issuer/issuer_instance.test.js` line 38 with the updated header value.

frederik:
@panva thanks for the patience. Can you please check the latest commit? The tests are passing now - I had to change a few more matchHeaders

PR #477, opened 2022-03-05
fix: client id type

liho00:
I found a bug AUTH_DISCORD_CLIENT_ID="replace with raw of number" return error, if i insert an alphabet between the number it works...this is due to discord client id is all pure number...

image

We should accept string && number as type for client_id as discord using pure numbers

image

Reference: https://github.com/directus/directus/issues/11956

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -174,7 +174,7 @@ class BaseClient {
     this.#issuer = issuer;
     this.#aadIssValidation = aadIssValidation;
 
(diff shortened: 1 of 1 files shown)


panva:
client_id has a string format. If it composes of all digit characters, that's fine, but pass it in as a string type and not a number.

liho00:
> client_id has a string format. If it composes of all digit characters, that's fine, but pass it in as a string type and not a number.

meaning directus package reading the client id as number and sending to openid-client... ok fine. thanks anyway....

PR #478, opened 2022-03-10
Verifying by `verification_uri` field variations on Device Authorization Flow.

devbarba:
I'm working with Azure OIDC and building the Device Authorization flow and was receiving this error: 
error-1
 
I did decide to do a deep verification what's happening and I did realize that Azure returns `verification_url` instead `verification_uri`:
print2

So, I did a verification before, regarding the field name and another fix on `expires_in` field verification.

### DIFF
--- a/lib/device_flow_handle.js
+++ b/lib/device_flow_handle.js
@@ -17,7 +17,9 @@ class DeviceFlowHandle {
   #maxAge;
   #response;
   constructor({ client, exchangeBody, clientAssertionPayload, response, maxAge, DPoP }) {
-    ['verification_uri', 'user_code', 'device_code'].forEach((prop) => {
+    const verificationUriField = response['verification_url'] ? 'verification_url' : 'verification_uri';
+
+    [verificationUriField, 'user_code', 'device_code', 'expires_in'].forEach((prop) => {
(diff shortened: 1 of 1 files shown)


panva:
Hi @devbarba,

Thank you for submitting a PR. I have reached out to my friends at Microsoft's Identity Standards team to let them know about this problem.

I understand Microsoft's implementation pre-dates any of the I-D or WG drafts, which is why these non-conformities are present. But, **it is not in the scope of this library to work around non conform behaviours.**

What I'll suggest to you is to make a PR to expose the DeviceFlowHandle class, which I'd accept, and make the requests to azure's `device_authorization_endpoint` yourself, then parse the response, modify it to be conform, instantiate DeviceFlowHandle manually and then proceed to `await handle.poll()` as usual.

panva:
I got a reply from Microsoft.

> The v2.0 endpoint (https://login.microsoftonline.com/{{tenant}}/oauth2/v2.0/devicecode) is RFC compliant.

Please use the latest interfaces from Azure and you're good to go.

devbarba:
> Please use the latest interfaces from Azure and you're good to go.

Great @panva, many thank's!!

PR #483, opened 2022-04-12
Add types for error constructors

mthadley:
I was recently writing some tests for handling errors from `openid-client` and while attempting to instantiate one of the errors, I noticed that the constructor types don't match the implementation.

Since [the types don't explicitly declare a `constructor`](https://github.com/panva/node-openid-client/blob/3adf1a47b5267f7ce20a7c0c29e7207c0ddf34ca/types/index.d.ts#L567-L589), they inherit they inherit it from `Error`, which doesn't match [their overloaded constructors](https://github.com/panva/node-openid-client/blob/3adf1a47b5267f7ce20a7c0c29e7207c0ddf34ca/lib/errors.js#L4).

So, I tried adding what I think are roughly the correct constructor types for the `OPError` and `RPError` classes. I saw that there is a `openid-client-tests.ts` file, so let me know if there's anything additional I should add there as well. Or, feel free to push any additional updates that you see fit to this branch.

Thanks!

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -8,6 +8,7 @@ import * as http2 from 'http2';
 import { URL } from 'url';
 import * as jose from 'jose';
 import * as crypto from 'crypto';
+import { format } from 'util';
 
 export type HttpOptions = Partial<
   Pick<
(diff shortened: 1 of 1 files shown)


panva:
Thank you @mthadley

PR #507, opened 2022-07-10
add support for node 18

Rugvip:
Node.js 18 is now halfway between it's initial release and becoming the active LTS, so it's pretty stable by now. There's no point blocking users of this library from using Node.js 18 right? You can get around it by ignoring engine checks during install, but that's an opt-out for all dependencies, which is pretty unfortunate. As long as none of the deprecation removals in 18 break anything in this library it's essentially free to start supporting 18 already.


panva:
As per https://github.com/panva/node-openid-client#install

> Node.js LTS releases Codename Erbium (starting with ^12.19.0) and newer LTS releases are supported. This means ^12.19.0 (Erbium), ^14.15.0 (Fermium), and ^16.13.0 (Gallium). Future LTS releases will be added to this list as they're released.

The next to be entering active LTS status will be Node.js 18 (Codename Hydrogen) in October 2022 as per the [release schedule](https://github.com/nodejs/release).

There were still deprecations and major change rollbacks during the very few 18.x releases that re-affirm me that I only wish to support LTS releases.

Rugvip:
There's a difference between only supporting LTS releases and only allowing LTS releases. Through the lens of typical package maintenance in the Node.js community your stance extremely strict.

panva:
I do not share your stance. 

We're talking about a 6 months time between april and october that even-numbered release reaches active LTS.

Most companies don't start upgrading their node version until it reaches EOL and mostly just go for the next one, not the recent one.

I'm perfectly fine having to ignore engines if I wish to run a piece of software outside of its intended runtime.

Rugvip:
Yep most companies don't start upgrading their node versions yet, but many also do. The maintenance cost for you is extremely minimal if there is any at all, all you need to do is to avoid using any new Node.js 18 APIs, which is ofc the case anyway since you support older versions too.

We have more than 3000 NPM dependencies in our project and out of all of those this is the only one that is blocking us from beginning to test towards and eventually support Node.js 18.

PR #511, opened 2022-07-25
fix: prevent redefinition of claims within tokenset

fredericoo:
Fixes a problem in which if `values` contained the property `claims`, it would replace the claims function. 

Discovered when trying to call `claims()` on `next-auth` via [this line](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/core/lib/oauth/callback.ts#L133).

## Checklist
✅ I've added tests to replicate the behaviour automatically

### DIFF
--- a/lib/token_set.js
+++ b/lib/token_set.js
@@ -3,6 +3,11 @@ const now = require('./helpers/unix_timestamp');
 
 class TokenSet {
   constructor(values) {
+    if (values && 'claims' in values) {
+      values['_claims'] = values.claims;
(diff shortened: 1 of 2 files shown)


panva:
Since there is no defined `claims` response parameter on the token, or authorization endpoints I don't agree with this PR's fix.

The solution needs to be ignore all unrecognized properties or throw them all in a single bag of "unrecognized", or something to that effect.

I'll ponder some more on this.

PR #485, opened 2022-04-12
chore: support Node 17

JalilArfaoui:
closes #484

PR #516, opened 2022-08-20
feat: add node 18 support

mstruebing:
This adds support and runs the GithubAction-Pipeline with node 18


panva:
As per #485 #507 and the [policy in the readme](https://github.com/panva/node-openid-client#install), this will be done in October with a v18.x version that is active LTS.

Rugvip:
@mstruebing I've published https://www.npmjs.com/package/openid-client-any-engine as a way around this for now. It's created by a script that downloads an existing `openid-client` version, tweaks the package.json, and publishes the modified package.

Issue #454, opened 2022-01-13
JWT not active yet, now 1642051718, nbf 1642051719: Id token Validation


sharmmoh1983:
While trying to validate the id token in callback , we are getting following error
JWT not active yet, now 1642051718, nbf 1642051719

I think   In id_token IDP  round the nbf up, but  open id connect  round the time down
Can i add  add 1 sec to make it work consistantly...

**To Reproduce**
Issuer and Client configuration: (inline or gist) - Don't forget to redact your secrets.
```js
// Issuer configuration (issuer.metadata) and how it is constructed (discovery or manual?)
import { Issuer } from 'openid-client';

const googleIssuer = await Issuer.discover('https://accounts.google.com');
console.log('Discovered issuer %s %O', googleIssuer.issuer, googleIssuer.metadata);
// Client configuration (client.metadata) and how it is constructed (fromUri or manual?)
const client = new googleIssuer.Client({
  client_id: '',
  redirect_uris: ['http://localhost:3000/cb'],
  response_types: ['id_token'],
  // id_token_signed_response_alg (default "RS256")
}); // => Client
```

Steps to reproduce the behaviour:

1. ```
 const client = new googleIssuer.Client({
  client_id: '',
  redirect_uris: ['http://localhost:3000/cb'],
  response_types: ['id_token'],
  // id_token_signed_response_alg (default "RS256")
}); // => Client
```
2.  ```
import { generators } from 'openid-client';
const nonce = generators.nonce();
// store the nonce in your framework's session mechanism, if it is a cookie based solution
// it should be httpOnly (not readable by javascript) and encrypted.
```

```
client.authorizationUrl({
  scope: 'openid email profile',
  response_mode: 'form_post',
  nonce,
});
```
3.  /```
/ assumes req.body is populated from your web framework's body parser
const params = client.callbackParams(req);
const tokenSet = await client.callback('https://client.example.com/callback', params, { nonce });
console.log('received and validated tokens %j', tokenSet);
console.log('validated ID Token claims %j', tokenSet.claims());
```

**Expected behaviour**
Id token getting validated successfully.

**Environment:**
 - openid-client version:^5.1.1
 - node version: v12

panva:
https://github.com/panva/node-openid-client/blob/main/docs/README.md#customizing-clock-skew-tolerance

Please refer to documentation and existing issues/discussions before opening an issue.

sharmmoh1983:
Thanks for the response .. Documentation was not clear enough to correlate the issue

PR #549, opened 2022-12-09
docs: update clock tolerance text

prust:
@panva: When searching the repo for a solution to `JWT not active yet`, the relevant docs did not appear in the search results. #454 did appear, which has a link to the appropriate docs, but this pull request makes the docs a little more discoverable by including the relevant error message.


panva:
Thank you @prust

PR #555, opened 2023-01-10
Import v8 from node internal

LoiLock:
Fixes:
image


panva:
There is no practical difference between `v8` and `node:v8` in the Node.js runtime. The former is also not supported in all major versions this library currently targets.

I'm going to go ahead and close this. I don't think there's anything to fix. If anything it might be a tooling issue related to your setup.

Issue #581, opened 2023-04-05
Client.userinfo does not accept other header than Authorization


thongxuan:
**Describe the bug**
This is not really a bug, but I think it is an important missing feature.
I'm working with Shopify OAuth, and they user 'X-Shopify-Access-Token' header for access token than 'Authorization'.

May be Shopify did not not correctly implement the OAuth2 specs, but if they do, then Client.userinfo function is missing a way to specify custom header for authentication. Currently there is option to specify params (to be sent as querystring).

**Expected behaviour**
Allow headers to be an option when calling Client.userinfo
```await client.userinfo(tokens, { headers: {'custom-header-here': 'access-token-here'} });```

**Environment:**
 - openid-client version: v5.3.0
 - node version: v18.13.0

**Additional context**

 - [x] the bug is happening on latest openid-client too.
 - [x] i have searched the issues tracker on github for similar issues and couldn't find anything related.

panva:
You can [customize individual http requests](https://github.com/panva/node-openid-client/blob/v5.x/docs/README.md#customizing-individual-http-requests) and add custom headers that way.

thongxuan:
Thank you for quick reply.

I think this allow per-request header config and not per-client config, so still good to have. 
We already have options to pass custom params, why not headers?

Using custom.http_options, we don't have header value to pass at the config time.

PR #582, opened 2023-04-05
feat(client): add ability to add custom header in Client.userinfo

thongxuan:
Allow specifying custom header in Client.userinfo, for OAuth server that does not accept access token in Authorization header.

This fixes: #581

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1165,7 +1165,7 @@ class BaseClient {
     return response;
   }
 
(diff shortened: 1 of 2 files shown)


panva:
https://github.com/panva/node-openid-client/issues/581#issuecomment-1497508833

> You can [customize individual http requests](https://github.com/panva/node-openid-client/blob/v5.x/docs/README.md#customizing-individual-http-requests) and add custom headers that way.

PR #583, opened 2023-04-17
Allow `default_max_age` to be optional

bestickley:
The [docs](https://github.com/panva/node-openid-client/blob/main/docs/README.md#new-clientmetadata-jwks-options) say that `default_max_age` is an optional Client constructor property. However, it is used as if it is a required field.
Currently `this.default_max_age` fails if `default_max_age` property is not set so this allows it to not be set.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -379,7 +379,7 @@ class BaseClient {
       params = await this.validateJARM(decrypted);
     }
 
-    if (this.default_max_age && !checks.max_age) {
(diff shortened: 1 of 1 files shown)


bestickley:
Alternatively, the TS types could be updated to make this property required in the `Issuer` constructor

panva:
> Currently this.default_max_age fails

How?

> this allows it to not be set.

This PR actually doesn't change anything, other than break compatibility with the supported nodejs versions.

PR #592, opened 2023-05-03
Update passport_strategy.js

Alankarsharma:
Bug fix.
If user opens two login page in same browser then state value is changed and login works from one tab only.


Alankarsharma:
Multiple login pages were opened because user clicked on login link in mail multiple times and once multiple tabs opened user closed all tabs except one. But because of this bug login work only from the last opened tab.

panva:
Hello @Alankarsharma 

> If user opens two login page in same browser then state value is changed and login works from one tab only.

They are two distinct authorization requests, they get two distinct states. If you want to work around this deficiency I'd suggest you to move the strategy into your own code and make the changes there.

> Bug fix.

It is not a bug.

PR #599, opened 2023-05-22
feat: rich errors in Issuer discovery / webfinger

sybereal:
Wrap errors during Issuer requests in `RPError` using `cause` to produce a complete stack trace.

As discussed in #598.


panva:
I believe every `request.call` call site should have its own reject handler with a non-generic message, wdyt?

panva:
fair enough

panva:
This has become stale, closing for now.

PR #607, opened 2023-06-29
Add example

andrewmcwatters:


andrewmcwatters:
Using Safari Version 16.5.1 (17615.2.9.11.7, 17615), https://github.com/panva/node-oidc-provider/blob/v8.2.2/example/views/login.ejs#L15 breaks when served from `localhost`.

Safari will attempt to redirect the the POST action to https.

Screen Shot 2023-06-28 at 9 32 16 PM

andrewmcwatters:
Switching both https://github.com/panva/node-oidc-provider/blob/v8.2.2/example/express.js and https://github.com/andrewmcwattersandco/node-openid-client/blob/feature/example/example/express.js to using HTTPS results in "too many redirects" when using Safari Version 16.5.1 (17615.2.9.11.7, 17615).

Screen Shot 2023-06-28 at 9 46 13 PM

andrewmcwatters:
I am unsure about how to continue. This pull requests demonstrates the "quick start" from [panva/node-openid-client](https://github.com/panva/node-openid-client#quick-start) using Express, and an updated Express example from https://github.com/panva/node-oidc-provider/pull/1224.

@panva Do you have any thoughts?

I suspect neither of these pull requests are desirable to merge, though https://github.com/panva/node-oidc-provider/pull/1223 might be.

I would like to run both of these codebases locally for development, but this behavior prevents me from doing so.

panva:
Hello @andrewmcwatters 

the problem with past examples in this repo was that they were meant to demonstrate the client capabilities but were mistaken for one size fits all copy pasteable code or that they were framework specific and spawned countless questions about "give me example for X", likely for the same "i want to copy paste a solution" reason. 

It is far more convenient for me to not have an example. My client modules are meant to be used by framework specific client modules written by oauth-minded folk who happen to be familiar with a given framework.

panva:
> I would like to run both of these codebases locally for development, but this behavior prevents me from doing so.

The examples are just that, examples, they may not work for everyone's desired setup. Given https://github.com/panva/node-oidc-provider/pull/1223 lands the express, koa, and standalone examples in oidc-provider are on-par feature wise.

They can be run locally on http, locally on https via NODE_ENV=production and ISSUER variables whilst behind a web server configured with valid certificates whos domain records point to a loopback, or on a hosting platform like heroku.

PR #637, opened 2023-11-23
refactor: issuer discovery

big-kahuna-burger:
Just splits uri detection code from a call to discovery endpoint and avoids duplication

### DIFF
--- a/lib/issuer.js
+++ b/lib/issuer.js
@@ -139,7 +139,7 @@ class Issuer {
 
   static async discover(uri) {
 
-    const wellKnownUri = getUri(uri);
+    const wellKnownUri = resolveWellKnownUri(uri);
 
     const response = await request.call(this, {
       method: 'GET',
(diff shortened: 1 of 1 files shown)


panva (changes requested):
please move the `getUri` function to the root of the file, so that it's not part of the `discover` function

lib/issuer.js
```diff
@@ -138,35 +138,8 @@ class Issuer {
...
-
-    let pathname;
-    if (parsed.pathname.endsWith('/')) {
-      pathname = `${parsed.pathname}.well-known/openid-configuration`;
-    } else {
-      pathname = `${parsed.pathname}/.well-known/openid-configuration`;
-    }
 
-    const wellKnownUri = url.format({ ...parsed, pathname });
+    const wellKnownUri = getUri(uri);
```
panva, line 142:
```suggestion
    const wellKnownUri = resolveWellKnownUri(uri);
```

lib/issuer.js
```diff
@@ -201,4 +174,19 @@ class Issuer {
   }
 }
 
+function getUri(uri) {
```
panva, line 177:
```suggestion
function resolveWellKnownUri(uri) {
```

panva (changes requested):

big-kahuna-burger:
@panva ptal

panva (approved):

PR #640, opened 2023-11-30
Add client credentials flow example to readme

sezanzeb:
As for the headline: names like "client credentials flow" "client credentials grant" and "client credentials grant flow" are flying around. I just went with the latter, idk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,8 +248,7 @@ but upon any other error it will reject. With tokenSet received you can throw aw
 
 ### Client Credentials Grant Flow
 
-Exchanging the client_id and client_secret of your application for an access token, to authenticate
-at a third-party api. Not on behalf of a user. Suitable for Machine-to-Machine authentication.
+Client Credentials flow is for obtaining Access Tokens to use with third party APIs on behalf of your application, rather than an end-user which was the case in previous examples.
 
 **See the [documentation](./docs/README.md#clientgrantbody-extras) for full API details.**
 
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ Client Credentials flow is for obtaining Access Tokens to use with third party A
 **See the [documentation](./docs/README.md#clientgrantbody-extras) for full API details.**
 
 ```js
-const client = new googleIssuer.Client({
+const client = new issuer.Client({
   client_id: 'zELcpfANLqY7Oqas',
   client_secret: 'TQV5U29k1gHibH5bx1layBo0OSAvAbRT3UYW3EWrSYBB5swxjVfWUa1BS8lqzxG/0v9wruMcrGadany3',
 });
--- a/README.md
+++ b/README.md
@@ -246,6 +246,27 @@ This will poll in the defined interval and only resolve with a TokenSet once one
 will handle the defined `authorization_pending` and `slow_down` "soft" errors and continue polling
 but upon any other error it will reject. With tokenSet received you can throw away the handle.
 
+### Client Credentials Grant Flow
+
+Exchanging the client_id and client_secret of your application for an access token, to authenticate
+at a third-party api. Not on behalf of a user. Suitable for Machine-to-Machine authentication.
+
+**See the [documentation][] for full API details.**
+
+```js
+const client = new googleIssuer.Client({
+  client_id: 'zELcpfANLqY7Oqas',
+  client_secret: 'TQV5U29k1gHibH5bx1layBo0OSAvAbRT3UYW3EWrSYBB5swxjVfWUa1BS8lqzxG/0v9wruMcrGadany3',
+  redirect_uris: ['http://localhost:3000/cb'],
+  response_types: ['code']
+});
+
+const tokenSet = await client.grant({
+  audience: 'insert-your-audience',
+  grant_type: 'client_credentials'
+});
+```
+
 ## FAQ
 
 #### Semver?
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -246,6 +246,27 @@ This will poll in the defined interval and only resolve with a TokenSet once one
 will handle the defined `authorization_pending` and `slow_down` "soft" errors and continue polling
 but upon any other error it will reject. With tokenSet received you can throw away the handle.
 
+### Client Credentials Grant Flow
+
+Exchanging the client_id and client_secret of your application for an access token, to authenticate
+at a third-party api. Not on behalf of a user. Suitable for Machine-to-Machine authentication.
+
+**See the [documentation][] for full API details.**
```
big-kahuna-burger, line 254:
```suggestion
**See the [documentation](./docs/README.md#clientgrantbody-extras) for full API details.**
```

README.md
```diff
@@ -246,6 +246,27 @@ This will poll in the defined interval and only resolve with a TokenSet once one
...
+```js
+const client = new googleIssuer.Client({
+  client_id: 'zELcpfANLqY7Oqas',
+  client_secret: 'TQV5U29k1gHibH5bx1layBo0OSAvAbRT3UYW3EWrSYBB5swxjVfWUa1BS8lqzxG/0v9wruMcrGadany3',
+  redirect_uris: ['http://localhost:3000/cb'],
+  response_types: ['code']
+});
+
+const tokenSet = await client.grant({
+  audience: 'insert-your-audience',
```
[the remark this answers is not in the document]
sezanzeb:
```suggestion
  // audience: 'optional-audience-name',
```

README.md
```diff
@@ -246,6 +246,27 @@ This will poll in the defined interval and only resolve with a TokenSet once one
...
 
+### Client Credentials Grant Flow
+
+Exchanging the client_id and client_secret of your application for an access token, to authenticate
+at a third-party api. Not on behalf of a user. Suitable for Machine-to-Machine authentication.
+
+**See the [documentation][] for full API details.**
+
+```js
+const client = new googleIssuer.Client({
```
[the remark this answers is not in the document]
panva: I agree with the proposed change.

README.md
```diff
@@ -246,6 +246,25 @@ This will poll in the defined interval and only resolve with a TokenSet once one
...
+**See the [documentation](./docs/README.md#clientgrantbody-extras) for full API details.**
+
+```js
+const client = new googleIssuer.Client({
+  client_id: 'zELcpfANLqY7Oqas',
+  client_secret: 'TQV5U29k1gHibH5bx1layBo0OSAvAbRT3UYW3EWrSYBB5swxjVfWUa1BS8lqzxG/0v9wruMcrGadany3',
+});
+
+const tokenSet = await client.grant({
+  audience: 'insert-your-audience',
```
panva, line 263:
```suggestion
  resource: 'urn:example:third-party-api',
```

sezanzeb: Looking into the auth0 web ui, the thing that jumps into view multiple times is the audience, and the quickstart shows it with an audience.

I tried to find something about this resource string, but couldn't. I also tried to look into the code, but the type of the grant function is just

```ts
export interface GrantBody {
  grant_type: string;

  [key: string]: unknown;
}
```

And resource is never mentioned in the docs of node-openid-client.

So I have no idea what the resource string is, and I definitely would want to see where to put the audience in the example.

sezanzeb: auth0 mentions an urn in the context of SAML: https://auth0.com/blog/url-uri-urn-differences/

the wikipedia entry says SAML is important for SSO: https://en.wikipedia.org/wiki/Security_Assertion_Markup_Language

Is this really relevant for the client credentials flow?

panva: https://datatracker.ietf.org/doc/html/rfc8707

big-kahuna-burger: @sezanzeb This repo's history will show you that it's aim is to support the OIDC Conforming behaviour as RP. 

That means couple of things:

- SHOULD promote conforming use cases, like suggesting RFC 8707 way of asking for a resource when those OPs implement RFC 8707.
- Should NOT promote any edge cases for different OP implementations around. 
Auth0 is OIDC provider vendor, but it's not conforming to 8707 spec as described. 
For example, I may not be using Auth0 as my OP, I may or may not be using `grant` function, and I don't want to be mislead into having to think about audience parameter at all in those cases.
So that's a rationale why `audience` should not land in docs in this library. More like **in your server implementation repo is where the place for those docs is.**
- Should NOT prevent you from passing extra parameters when those are required. It's also stated in types with this type descriptor: `[key: string]: unknown;`, that's why typescript is not complaining when you add audience. Also, you can look [at source code ](https://github.com/panva/node-openid-client/blob/main/lib/client.js#L1341) and see that all arguments you define are in fact being passed to provider.

sezanzeb: > Should NOT promote any edge cases for different OP implementations around.

This is for giving hints for less knowledgeable people like me. I'm not a specialist for oauth, and my intention for using a library is, that it will help me set things up correctly without having to have a deeper understanding of the thing at hand.

Actually, I don't remember where I even got the idea that I can insert audience there. It might have been that I just confused `AuthorizationParameters` with `DeviceAuthorizationParameters` when looking into the code, and therefore by pure luck correctly inserted `audience` into the `grant` function...

I made another commit, this time removing audience from README.md, and saying that audience is non-standard in docs/README.md

docs/README.md
```diff
@@ -381,7 +381,8 @@ Performs an arbitrary `grant_type` exchange at the `token_endpoint`.
 
 - `body`: `<Object>`
   - `grant_type`: `<string>`
-  - other properties may be provided depending on the grant in question
+  - other standard properties like `resource` (RFC 8707), or non-standard properties like `audience` may be provided
```
panva, line 384:
```suggestion
  - other properties may be provided depending on the grant in question
```

PR #658, opened 2024-01-31
feat: expose AAD_MULTITENANT on Issuer

charsleysa:
Expose the AAD_MULTITENANT on the Issuer class as a static property. This allows creating instances of Issuer by calling the constructor and supplying the metadata object directly instead of needing to execute `await Issuer.discover(...)`.

### DIFF
--- a/lib/issuer.js
+++ b/lib/issuer.js
@@ -80,6 +80,10 @@ class Issuer {
     return clone(Object.fromEntries(this.#metadata.entries()));
   }
 
+  static get AAD_MULTITENANT() {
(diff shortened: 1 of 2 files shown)


panva:
Hello @charsleysa,

this is [intentionally](https://github.com/panva/node-openid-client/pull/148#issuecomment-483348258) only through discovery.

charsleysa:
Thanks for the information @panva. We've migrated to using `oauth4webapi` instead and followed your advice provided in one of the issues https://github.com/panva/oauth4webapi/issues/100#issuecomment-1890380232

PR #660, opened 2024-02-07
refactor: avoid use of prototype attributes in keystore queries

pnappa:
When this library is run under environments which prohibit use of the deprecated `__proto__` getter/setter, it may cause a crash due to reading from the __proto__ property. An example scenario is when the `--disable-proto=throw` flag is used for NodeJS.

To disable consideration of the `prototype`, `__proto__`, and `constructor` fields fields when hashing, the `respectType` flag can be toggled off. This is not a problem for this library, as we are querying only for the existence of a object created via an object literal, which will have these properties all the same anyway.

### DIFF
--- a/lib/helpers/issuer.js
+++ b/lib/helpers/issuer.js
@@ -75,6 +75,7 @@ async function queryKeyStore({ kid, kty, alg, use }, { allowMulti = false } = {}
     ignoreUnknown: true,
     unorderedArrays: true,
     unorderedSets: true,
+    respectType: false,
   });
 
   // refresh keystore on every unknown key but also only upto once every minute


pnappa:
If you're curious as to why I'm using that Node flag, I'm looking to migrate a codebase to Deno (which has removed the __proto__ feature), and it is necessary to complain very loud about potential scenarios where it would lead to silent incompatibility. :)

panva:
@pnappa thank you Patrick, though please note that this is a [Node.js module](https://github.com/panva/node-openid-client#how-do-i-use-it-outside-of-nodejs) and compatibility with Deno or other runtimes is not a goal. On the other hand https://github.com/panva/oauth4webapi is built using just Web runtime APIs, it is however a much lower level API.

PR #674, opened 2024-04-17
Bumped lru cache to version 10

SamTV12345:
This should also resolve the issue we have with the different lru-cache dependencies:

https://github.com/ether/etherpad-lite/issues/6330

All tests are passing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,7 @@ npm-shrinkwrap.json
 
 # TS test
 types/*.js
+
+
+# Editors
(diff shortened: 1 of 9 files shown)


panva:
I cannot bump the version of lru-cache without dropping support for older Node.js versions. That can only be done in semver-major update.

I will obviously update all modules to their latest versions when I get the time to do a major release in the future but I don't have the capacity to finalize my plans for v6.x just yet.

PR #675, opened 2024-04-24
fix: check key.jwk.kid in keystore when comparing

xoapit:


panva:
Hi @xoapit 

can you explain your fix? What is it fixing, what conditions to reproduce, add a regression test.

xoapit:
When integrating with ID.me, I found that the **key.jwk** from IdP does not include **kid**. We should only check kid if it is defined in IdP, if not, we should skip it.

PR #676, opened 2024-05-23
feat: add resetHttpOptionsDefaults for unsetting defaults

kendlee:
Problem: Whenever `custom.setHttpOptionsDefaults` is called to set headers, there are no available method to remove/unset any of the added headers. This becomes problematic on the event that multiple Client instances need to be created, where the other doesn't need the preconfigured headers.

Currently, the only means to unset headers will be to override them is to call `custom.setHttpOptionsDefaults`, and set the same header key with a null value.

Proposal is to add a resetHttpOptionsDefaults method which unsets everything to the defaults, so that there is no need to unset any value explicitly.

Pseudo-code on usage:
```javascript
custom.setHttpOptionsDefaults({
  headers: {
    Origin: 'some-origin', // required by microsoft provider
  }
});
const msclient = new issuer.Client(config);
// some code to retrieve refreshToken
const tokens = await msclient.refresh(credential.refreshToken); // uses header while performing request

// unless this is reset, all requests will include the preconfigured header
custom.resetHttpOptionsDefaults();

const anotherclient = new issuer.Client(anotherConfig);
```

### DIFF
--- a/lib/helpers/request.js
+++ b/lib/helpers/request.js
@@ -14,7 +14,15 @@ const pick = require('./pick');
 const { deep: defaultsDeep } = require('./defaults');
 const { HTTP_OPTIONS } = require('./consts');
 
-let DEFAULT_HTTP_OPTIONS;
+const DEFAULT_HTTP_OPTIONS = {
+  headers: {
+    'User-Agent': `${pkg.name}/${pkg.version} (${pkg.homepage})`,
+    'Accept-Encoding': 'identity',
+  },
+  timeout: 3500,
+};
+
+let storedHttpOption;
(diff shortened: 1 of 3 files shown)


panva:
Hi @kendlee,

I recognize the many issues that come from default headers being configured on a library level.  

That's why all options can be [configured](https://github.com/panva/node-openid-client/blob/v5.x/docs/README.md#customizing-individual-http-requests) on a more granular bases as well. The Issuer constructor has the same issue as default library level though...

These options will be removed entirely in the next major in favour of a different API (similar to [got.extend](https://github.com/sindresorhus/got/blob/v14.3.0/documentation/10-instances.md#gotextendoptions-instances)).

PR #713, opened 2024-10-21
fix: in case there is a port number other than 80 or 443

mihui:
```javascript
const strategy = new Strategy({
  config: myConfig, scope: myScope,
  callbackURL: 'https://localhost:8080/callback',
}, myVerifyFunc);

// serializeUser, deserializeUser, etc...

passport.use('my-strategy', strategy);
```

When the callback request reach to the callback URL, it says:
```shell
CSIAQ0167E The redirection URI that is provided in the request [https://localhost/callback] is either invalid, or does not meet the matching criteria for the registered redirection URI.
```

Apparently the port number is missing...

### DIFF
--- a/src/passport.ts
+++ b/src/passport.ts
@@ -385,7 +385,17 @@ export class Strategy implements passport.Strategy {
    *   callbackURL was specified in the Strategy constructor
    */
   currentUrl(req: express.Request): URL {
-    return new URL(`${req.protocol}://${req.host}${req.url}`)
+    const DEFAULT_PORTS = [80, 443, 0];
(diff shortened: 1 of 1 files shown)


panva:
Are you using passport@5?

panva:
Are you using express@5? Because AFAICT `req.host` is fixed in express 5 and properly returns port.

If you don't use the latest express version where this is fixed or use some other non-standard setup - you can just overload the `currentUrl` method so that it works for your conditions, that's the whole reason why this method is public and typed - so that you can control it.

PR #774, opened 2025-04-02
Uncaught bad http response in method authorizationCodeGrant fix

wjes:


panva:
This is handled inside the followup method that processes a Response object, namely `processAuthorizationCodeResponse` from oauth4webapi.

wjes:
🤦🏻‍♂️ sorry, my bad

Issue #797, opened 2025-06-17
[Bug] `selectCryptoKeyForDecryption` passes over valid key because it doesn't check `kid` correctly


zackchengyk:
### What happened?

Suppose you have a key `myEncKey: CryptoKey` which has a `kid`.

If you set `enableDecryptingResponses` with that key, if you try to decrypt a JWE that was encrypted using that key, the key won't get correctly selected by `selectCryptoKeyForDecryption` during attempted decryption.

The cause of this bug is easier to understand by looking at the source code:

1. Function [`enableDecryptingResponses`](https://github.com/panva/openid-client/blob/29db5540de3345b525d014b65e6967eb4aa0bbb7/src/index.ts#L1492C17-L1492C42) normalizes all the keys (`CryptoKey | DecryptionKey`) it receives.
2. Notably, if it receives `myEncKey: CryptoKey`, it'll normalize it by wrapping it into `{ key: myEncKey }` [here](https://github.com/panva/openid-client/blob/29db5540de3345b525d014b65e6967eb4aa0bbb7/src/index.ts#L1524C7-L1524C24) without setting `kid` or `alg` (it goes into this branch because `CryptoKey`s don't have a `key` field). Now, `key.kid` is `undefined`.
3. Later, function [`decrypt`](https://github.com/panva/openid-client/blob/29db5540de3345b525d014b65e6967eb4aa0bbb7/src/index.ts#L1647C16-L1647C23) calls `compactDecrypt` using a key selected by `selectCryptoKeyForDecryption` out of the 'normalized' keys from earlier.
4. `selectCryptoKeyForDecryption` runs through the 'normalized' keys, but the [first check it does](https://github.com/panva/openid-client/blob/29db5540de3345b525d014b65e6967eb4aa0bbb7/src/index.ts#L1617C5-L1619C6) is always `kid === key.kid`. This causes the correct key to be skipped, because `key.kid` is `undefined`.

The fix is simple: just do a `key.kid !== undefined` check in the same line. I've tested it and this works because we then fall through to the [`checkCryptoKey`](https://github.com/panva/openid-client/blob/29db5540de3345b525d014b65e6967eb4aa0bbb7/src/index.ts#L1625C12-L1625C26) function which works as intended.

[...]

PR #798, opened 2025-06-17
fix: add `undefined` check in `selectCryptoKeyForDecryption`

zackchengyk:
Resolves issue [#797](https://github.com/panva/openid-client/issues/797).

- Adds an `undefined` check for `key.kid`, and
- Makes the `undefined` check explicit for `key.alg`.


zackchengyk:
Closed as issue #797 was closed

PR #817, opened 2025-08-12
Allow passing URLSearchParams directly to implicitAuthentication

nwalters512:
Given that the function is already using `URLSearchParams` internally, it seems like a natural extension to allow one to pass an instance of `URLSearchParams` if they already have it. This makes `implicitAuthentication` usable even when one doesn't have a web `Request` object.

### DIFF
--- a/docs/functions/implicitAuthentication.md
+++ b/docs/functions/implicitAuthentication.md
@@ -25,7 +25,7 @@ Response.
 | Parameter | Type | Description |
 | ------ | ------ | ------ |
 | `config` | [`Configuration`](../classes/Configuration.md) | - |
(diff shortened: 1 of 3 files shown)


panva:
Hello @nwalters512

I'm not sure in what case the existing inputs aren't "usable". The [examples](https://github.com/panva/openid-client/blob/v6.x/docs/functions/implicitAuthentication.md#examples) cover the use cases I have in mind.

nwalters512:
"When using a form_post response mode without a Request instance" from those docs matches my scenario. While that works, it's simply very clunky to require one to construct a full URL. It seems especially silly given that `implicitAuthentication` will immediately unpack things back into a `URLSearchParams` instance. If I already have a `URLSearchParams`, it seems like I should just be able to pass it directly.

I should have said this PR makes it _more_ usable (or rather, more ergonomic).

